PR #77, opened 2014-05-02
Adding firewall support

blairham:
This will add some basic firewall support to the windows cookbook. This does not include all the features but a good subset.

### DIFF
--- a/providers/firewall.rb
+++ b/providers/firewall.rb
@@ -22,7 +22,7 @@ include Chef::Mixin::ShellOut
 
 action :add do
   unless @new_resource.created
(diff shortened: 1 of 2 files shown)


blairham:
@juliandunn or @btm  I had made the changes requested a while ago but this was close. can we re-open this?

jaym:
There is a cookbook dedicated to windows firewall: https://github.com/lynx44/windows_firewall

juliandunn:
@blairham @jaym Yeah, I'd personally rather invest the time in the `windows_firewall` cookbook rather than adding more primitives to the Windows cookbook at this point. We're trying to excise the Windows cookbook from being a pre-req on Windows at all.

PR #113, opened 2014-09-10
Update scheduled tasks if definition changed

gturri:
With this patch, when we update the command or the user, and if we set
"force" to true, then the scheduled task is updated.

BTW, I see in https://github.com/opscode-cookbooks/windows/blob/master/CONTRIBUTING that I should open a Jira task to contribute. However, according to http://www.getchef.com/blog/2014/06/16/chef-issue-tracking-and-contribution-workflow-updates/ I believe that this file is outdated. Please feel free to tell me If I'm wrong.

### DIFF
--- a/providers/task.rb
+++ b/providers/task.rb
@@ -22,9 +22,17 @@ require 'chef/mixin/shell_out'
 include Chef::Mixin::ShellOut
 
 action :create do
+  has_changed = nil
   if @current_resource.exists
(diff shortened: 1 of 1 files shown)


smurawski:
@gturri Thanks for the PR!  I've implemented this slightly differently, so the functionality you described should be on master now.  I'll cut a new release of this cookbook on 5/14/2015.

PR #117, opened 2014-09-22
Jdmundrawala/chef update

jaym:
A stab at a recipe that can update chef


smurawski:
Incomplete contribution.  @jaym feel free to resurrect, but I think the Chef Upgrade RFC work will supercede this. https://github.com/chef/chef-rfc/blob/master/rfc070-chef-upgrade.md

PR #122, opened 2014-09-25
Move reboot_pending? method to windows_helper

Annih:
Move method `reboot_pending?` from `windows_reboot_handler` to `windows_helper`.
This allows thirdparty providers to use the detection code.
I also added another check to the `reboot_pending?` method to handle CBS required reboot.


Annih:
I think you are right @btm, the Chef DSL is way better, so I don't need to move the method out of the Handler anymore.
Therefore I created a new pull request #236  to fix some bugs and use the DSL when possible.

PR #132, opened 2014-10-21
conform to new chefspec matcher definition syntax for versions of chefspec >= 4.1

mwrock:
chefspec 4.1.0 and greater displays deprication messages when using the `ChefSpec::Runner.define_runner_method` syntax. This updates these to use the new `define_matcher` but should not break those using the older chefspec gem.


kowal:
This fix is included in versions 1.36.1, which was released to supermarket.

PR #143, opened 2014-11-11
Fixes Kernel::WindowsRebootHandler

EasyAsABC123:
This loads the namespace after the file is loaded to the server:
Error Message: undefined constant Kernel::WindowsRebootHandler


ksubrama:
We don't actually need this here because chef_handler should now have a proper matcher and work with ChefSpec.  There was an issue because https://github.com/opscode-cookbooks/chef_handler/tree/master/libraries had "matcher.rb" and "matchers.rb" but ChefSpec only looks at the latter.  The former was present in the last release.  I'm fixing this up and cutting a new release which should smooth things out.

EasyAsABC123:
@ksubrama good to know the details on why this was an issue and that it is fixed.

ksubrama:
Well...  I turns out that was one issue...  but I don't know if it was the only one.  I noticed another bug where a previously loaded class is not actually unloaded by chef_handler.  I should really stop trying to find bugs.  I keep making software that previously worked stop working by looking at it.

EasyAsABC123:
That's definitely a skill, add it to your LinkedIn I'll endorse you ;)

Justin Schuhmann

ksubrama:
https://github.com/opscode-cookbooks/chef_handler/pull/20

ksubrama:
#216 removes the workaround now that chef_handler 1.1.9 is out.

PR #152, opened 2014-12-09
Fix documentation for windows_task

Tech356:
Currently the documentation states that when a user is given, it also needs a password.

    user: The user to run the task as. (requires password)

This is incorrect. The [code](https://github.com/opscode-cookbooks/windows/blob/master/providers/task.rb#L28) only requires that a user is supplied if a password is supplied. If a user is supplied, a password is not required.

    if @new_resource.user and @new_resource.password.nil? then Chef::Log.debug "#{@new_resource} did not specify a password, creating task without a password" end

This is useful when setting the user as the system account since the system account has no password.

    windows_task "Run Foobar Daily" do
        command 'foobar.exe'
        frequency :daily
        cwd 'c:\windows\system32'
        user 'NT AUTHORITY\SYSTEM'
    end

### DIFF
--- a/README.md
+++ b/README.md
@@ -488,7 +488,7 @@ Server 2008 due to API usage.
 - name: name attribute, The task name.
 - command: The command the task will run.
 - cwd: The directory the task will be run from.
-- user: The user to run the task as. (requires password)
+- user: The user to run the task as.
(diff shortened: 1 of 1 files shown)


Tech356:
Fixed by #204

PR #162, opened 2015-01-11
Adding basic rspec tests for win2k12r2

spudnic:
I find that developing cookbooks with server spec only is way to slow. Each time you run kitchen test it can take several minutes if not more, especially if there are multiple platforms.  I think getting some basic rspec tests that only take a few seconds really helps to speed things up.  Since 8.1 windows is kind of old I went with versions 2008r2 and 2012R2.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,5 @@
 Berksfile.lock
 Gemfile.lock
 .bundle
+.rvmrc
+.idea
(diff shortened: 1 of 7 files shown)


spudnic:
I have signed the paper work and linked to super market.

ksubrama:
I believe an equivalent to this change has already been merged into master.  Thanks for contributing!

PR #163, opened 2015-01-13
Add support for the Microsoft Update Standalone Installer (MSU) packages

zl4bv:


mwrock:
closing since `windows_package` has moved to chef core but we definitely would like to add MSU support there and PR's are welcome.

zl4bv:
Thanks @mwrock!

PR #164, opened 2015-01-15
Only define ExpandEnvironmentStrings when not already defined

maoueh:
Similar to what was done in https://github.com/opscode-cookbooks/windows/pull/25, only define `ExpandEnvironmentStrings` when it is not defined already.

I wasn't sure about freezing it since it's not a scalar constant, hence I chose to not freeze it.

This solves already defined constant warning when cookbook is included multiple times:

```
cookbooks/windows/libraries/windows_helper.rb:29: warning: already initialized constant Windows::Helper::ExpandEnvironmentStrings 
cookbooks/windows/libraries/windows_helper.rb:29: warning: previous definition of ExpandEnvironmentStrings was here
```

Regards,
Matt


tyler-gray:
We are seeing the warnings as well with output here:
https://gist.github.com/tyler-gray/ee09af66072be99f6983

ksubrama:
:+1: Thanks!

PR #166, opened 2015-01-15
Separated the installed_packages function

EasyAsABC123:
this allows for all providers to test if something is installed.  As well as external usage by any other recipe or provider anywhere.


EasyAsABC123:
moving to Windows::Helper

PR #167, opened 2015-01-16
adding a certificate provider to install ssl certs into the store

blairham:


Tech356:
Do you have some examples on how to use this? Maybe update the README.md with description, usage and examples. 

From reading the code it looks like you are creating a new store for each certificate instead of adding the certificates to an existing store.  Is there a reason for this design?

blairham:
Yes I'll update the README.md

As for the store, it's adding it to the current store and you are specifying the StoreName. This was obtained from here: http://ss64.com/nt/certutil.html

providers/certificate.rb
```diff
@@ -0,0 +1,63 @@
...
+    shell_out!(cmd)
+    new_resource.updated_by_last_action true
+    Chef::Log.info("#{@new_resource.name} removed from store.")
+  else
+    Chef::Log.debug("#{@new_resource.name} does not exist in store - nothing to do.")
+  end
+end
+
+def load_current_resource
+  @current_resource = Chef::Resource::WindowsService.new(@new_resource.name)
```
Tech356: Shouldn't this be `WindowsCertificate` instead of `WindowsService`?

resources/certificate.rb
```diff
@@ -0,0 +1,30 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+actions :add, :delete
+
+attribute :store_name, :kind_of => String, :name_attribute => true
+attribute :cert_id, :kind_of => String
+attribute :infile, :kind_of => String
```
Tech356: I would prefer this to use `:infile` as the name attribute. I think there are more use cases where the certificate path would be unique.

providers/certificate.rb
```diff
@@ -0,0 +1,63 @@
...
+  end
+end
+
+private
+
+def cert_exists?(name)
+  Chef::Log.debug "Checking to see if this cert is in the store: '#{ name }'"
+  cmd = shell_out!("certutil -verifystore #{name}", {:returns => [0,-2147024894]})
+  cmd.stderr.empty? && (cmd.stdout =~ /Certificate is valid/i)
+end
```
Tech356: This validates the certificate instead of just checking that it exists. It fails on certificates that don't have a full chain, like self-signed. I think it would be better to use the Serial Number or Hash to check if the certificate exists. For instance hash the infile with `certutil -hashfile <infile>` and see if that can be found in the store.

providers/certificate.rb
blairham: Good catch, you are correct.

README.md
```diff
@@ -90,6 +90,35 @@ windows_batch 'echo some env vars' do
...
+windows_certificate 'example.com' do
+  infile: C:/certificates/example.com.pem
+end
+```
+
+Delete certificate from store
+```ruby
+windows_certificate 'example.com' do
+  action :delete
+  store_id: "089D8C1FD45893D7BF76F3788D3B1ED7E0974100"
```
Tech356: I assume you mean `cert_id` instead of `store_id` here?

resources/certificate.rb
blairham: sure then if you don't specify a store_name it would use the path? I could see that as acceptable.

Tech356: Now I am not so sure. There isn't really a good unique element to use as the name. The store name is used in both adding and deleting where as the infile is just used in adding and cert_id is used in just deleting. 

The main problem I have with using store name as the name attribute I when adding multiple certificates to a store. I have a recipe where I am adding 3 certificates the TrustedPublisher store. In that case couldn't use the name attribute anyway since it would cause resource cloning. 

Adding certificates with `store_name` as `name_attribute`

    windows_cerificate 'TrustedPublisher 1' do
        action :add
        store_name 'TrustedPublisher'
        infile 'mycert.cer'
    end

    windows_cerificate 'TrustedPublisher 2' do
        action :add
        store_name 'TrustedPublisher'
        infile 'mycert2.cer'
    end

Adding certificates with `infile` as `name_attribute`

    windows_cerificate 'mycert.cer' do
        action :add
        store_name 'TrustedPublisher'
    end

    windows_cerificate 'mycert2.cer' do
        action :add
        store_name 'TrustedPublisher'
    end

Deleting certificates with `store_name` as `name_attribute`

    windows_cerificate 'TrustedPublisher 1' do
        action :delete
        store_name 'TrustedPublisher'
        cert_id '123456789'
    end

    windows_cerificate 'TrustedPublisher 2' do
        action :delete
        store_name 'TrustedPublisher'
        cert_id '123456789'
    end

Deleting certificates with `infile` as `name_attribute`

    windows_cerificate 'TrustedPublisher 1' do
        action :delete
        store_name 'TrustedPublisher'
        cert_id '123456789'
    end

    windows_cerificate 'TrustedPublisher 2' do
        action :delete
        store_name 'TrustedPublisher'
        cert_id '123456789'
    end

providers/certificate.rb
```diff
@@ -0,0 +1,63 @@
...
+# limitations under the License.
+#
+
+require 'chef/mixin/shell_out'
+
+include Chef::Mixin::ShellOut
+
+action :add do
+  unless @current_resource.exists
+    cmd = "certutil -f -addstore #{@new_resource.name} #{@new_resource.infile}"
```
Tech356: `infile` should be validated as a valid path before it is used.

providers/certificate.rb
blairham: So I've been looking into this and this actually verify the cert in the store, so if it's not in the store it will not return "Certificate is valid". I haven't been able to identify another way that doesn't open a GUI window. I'm open to suggestions, I do like the idea of hashing it and finding it but is seems that verifystore is already doing that.

https://technet.microsoft.com/library/cc732443.aspx#BKMK_verifystore

Tech356: What if there are multiple certificates in a single store? Run `certutil -verifystore ca` or `certutil -verifystore root` and you'll see what I mean. On my machines the `ca` store has 3 certificates and only 1 has "Certificate is valid". In the `root` store I have 11 certificates and only 7 have "Certificate is valid". 

If you pass `cert_id` to the verifystore command, that solve the issue with multiple certificates. But it does not solve issues with non-valid certificates. I have a couple use cases where I am installing software that is so old that the signing certificate has expired. To be able to install the software without a GUI prompt I have to add the certificate to the certificate manager before I install the software. When I run `-verfystore` over that certificate it returns "Expired certificate".

resources/certificate.rb
```diff
@@ -0,0 +1,30 @@
...
+
+actions :add, :delete
+
+attribute :store_name, :kind_of => String, :name_attribute => true
+attribute :cert_id, :kind_of => String, :required => true
+attribute :infile, :kind_of => String
+
+attr_accessor :exists
+
+def initialize(name, run_context=nil)
```
ksubrama: You can use default_action now without having to override initialize.

resources/certificate.rb
```diff
@@ -0,0 +1,30 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+actions :add, :delete
+
+attribute :store_name, :kind_of => String, :name_attribute => true
+attribute :cert_id, :kind_of => String, :required => true
```
ksubrama: cert_id is required but the example in the readme doesn't require one.

providers/certificate.rb
```diff
@@ -0,0 +1,68 @@
...
+  if cert_exists?(@new_resource.name,@new_resource.cert_id)
+    @current_resource.exists = true
+  end
+end
+
+private
+
+def cert_exists?(name,id)
+  Chef::Log.debug "Checking to see if this cert is in the store: '#{ name }'"
+  cmd = shell_out!("certutil -store #{name} #{id}", {:returns => [0]})
```
ksubrama: This makes me feel uncomfortable given just how many things certid could be.  If certId is 0, then all it does it verify if certificate 0 exists.  That seems...  not right.

ksubrama:
#212 and this attack the same problem and has some of the same issues.  I'll reconcile these and merge it in manually.

ksubrama:
Thanks for taking the time to contribute code!

PR #170, opened 2015-01-24
powershell_out uses powershell 64bit on x64 node

### LINKED ISSUE #169: powershell_out use 32-bit powershell on x64 node
Hi,

Recently I encountered some weird errors with my powershell scripts and I finally discovered that `powershell_out` just run `powershell [flags] [script]`.
At first look, it seems correct... but because currently chef is running with a 32-bit ruby this is wrong!

I found out that the module `Powerhell::Helper` has an `interpreter` function doing what we need :)

Annih:
This PR resolve #169 
It includes `Powershell::Helper` inside `Chef::Mixin::PowershellOut` and leverage the [`interpreter` function](https://github.com/opscode-cookbooks/windows/blob/master/libraries/powershell_helper.rb#L31)

### DIFF
--- a/libraries/powershell_out.rb
+++ b/libraries/powershell_out.rb
@@ -2,6 +2,7 @@ class Chef
   module Mixin
     module PowershellOut
       include Chef::Mixin::ShellOut
+      include Powershell::Helper
 
       begin
         include Chef::Mixin::WindowsArchitectureHelper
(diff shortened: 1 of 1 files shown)


Annih:
#169 is not a real issue.
The file system redirection is disabled.

PR #172, opened 2015-01-28
Add base serverspec test

mattstratton:
Added a base serverspec test to force a busser test to allow for Kitchen verify. Also changed the Chef client version from 12.0.0 to "12" so it just pulls the latest version of chef-client 12.

### DIFF
--- a/.kitchen.yml
+++ b/.kitchen.yml
@@ -31,7 +31,7 @@ platforms:
       box: windows-8.1
       box_url: https://s3-us-west-2.amazonaws.com/chef-box-cutter/windows/virtualbox4.3.20/eval-win81x64-enterprise-nocm-1.0.3.box
     provisioner:
-      require_chef_omnibus: 12.0.0
+      require_chef_omnibus: 12
 
 suites:
   - name: default
--- a/.kitchen.yml
+++ b/.kitchen.yml
@@ -17,19 +17,19 @@ platforms:
   - name: windows-8.1-chef11.10.4
     driver_config:
       box: windows-8.1
-      box_url: https://s3-us-west-2.amazonaws.com/chef-box-cutter/windows/virtualbox4.3.20/eval-win81x64-enterprise-nocm-1.0.3.box
+      box_url: http://cdn.box-cutter.com/windows/virtualbox4.3.20/eval-win81x64-enterprise-nocm-1.0.2.box
     provisioner:
       require_chef_omnibus: 11.10.4
   - name: windows-8.1-chef11.16.4
     driver_config:
       box: windows-8.1
-      box_url: https://s3-us-west-2.amazonaws.com/chef-box-cutter/windows/virtualbox4.3.20/eval-win81x64-enterprise-nocm-1.0.3.box
+      box_url: http://cdn.box-cutter.com/windows/virtualbox4.3.20/eval-win81x64-enterprise-nocm-1.0.2.box
     provisioner:
       require_chef_omnibus: 11.16.4
   - name: windows-8.1-chef12
     driver_config:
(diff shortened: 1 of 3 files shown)


jdmundrawala:
serverspec!? we are pester people.

but still, awesome

Delete the commented out line in the gemfile.
Other than that, :shipit:

mattstratton:
LOL - I wanted to test out busser. We can add a pester test next. 

I removed the commented out bits and bobs.

tyler-ball:
This worked for me locally on a 2012r2 image so :+1:

.kitchen.yml
```diff
@@ -31,7 +31,7 @@ platforms:
       box: windows-8.1
       box_url: http://cdn.box-cutter.com/windows/virtualbox4.3.20/eval-win81x64-enterprise-nocm-1.0.2.box
     provisioner:
-      require_chef_omnibus: 12.0.0
+      require_chef_omnibus: 12
```
tyler-ball: One downside of this I just came across - the PS script which [detects the currently installed](https://github.com/test-kitchen/test-kitchen/blob/windows-guest-support/support/chef_helpers.ps1#L6-L20) version only does direct string matching.  So putting a `12` here instead of `12.0.3` reinstalls the omnibus many times on an existing client.

mattstratton:
@tyler-ball That's a bummer. @fnichol thought it would work, but that's probably because *nix doesn't use the PS script.

I will say that forcing it to 12.0.0 breaks it. I'm okay with changing it to 12.0.3, but we should figure out how to make it act like non-windows in this regard.

.kitchen.yml
jdmundrawala: I think we should fix this. We can ask omnitruck for the one it thinks is the best match, and do a comparison off of that

tyler-ball: I am a PS newb - can we look at that together?  Or would you rather I file a separate issue?

jdmundrawala: yep.

cmluciano: Would setting it to true be ok for now? I took a look at the omnitrunk API docs. I was pondering how to grab the platform info to send to it. Will keep digging.

mattstratton: Do we want to wait for this to be fixed in test-kitchen before merging this? Seems logical.

jdmundrawala: No. The nix stuff behaves the same way iirc

cmluciano: Ya it seems pretty similar https://github.com/test-kitchen/test-kitchen/blob/master/support/chef_helpers.sh

cmluciano:
This might be outside the scope of this PR, but I was wondering if we might be able to change the vm.hostname under here to something that is not false. Some of my custom LWRPs are choking on this. 

Was there a specific reason to name it this that anyone can recall? I know windows prefers names <15 characters.

mattstratton:
Sorry for getting confused here - can I get some +1 on merging this as it stands, or should I revert it back to explicitly calling 12.0.03?

jdmundrawala:
:+1:

tyler-ball:
:+1:

btm:
@cmluciano can you file a separate issue about that? We're talking about it in a standup right now.

cmluciano:
@btm Sure. I can open a PR too if we have an idea on a name for this vm.

.kitchen.yml
btm: This should be fixed in windows-guest-support by: https://github.com/test-kitchen/test-kitchen/pull/590

https://github.com/test-kitchen/test-kitchen/commit/b74a96e0add86d6cf1011b91437a55274785dd62

PR #178, opened 2015-02-08
Add icon support windows_shortcut

lambacck:


lambacck:
Fixes #177

Tech356:
I know you didn't write the resource originally but would you be willing to add documentation to the README? At the moment there is nothing in the README about the windows_shortcut resource.

lambacck:
@Tech356 I added some basic docs for windows_shortcut

README.md
```diff
@@ -459,6 +459,34 @@ windows_registry 'HKCU\Software\Test' do
...
+
+Add a shortcut all users desktop:
+```ruby
+require 'win32ole'
+all_users_desktop = WIN32OLE.new("WScript.Shell").SpecialFolders("AllUsersDesktop")
+
+windows_shortcut "#{all_users_desktop}/Notepad.lnk" do
+    target "C:\\WINDOWS\\notepad.exe"
+    description "Launch Notepad"
+    iconlocation "C:\\windows\\notepad.exe, 0"
```
btm: The string should contain a fully qualified path and an index associated with the icon. See example for more details.

Should the index ever be non-zero?

lambacck: Yes, executables and .ico files can have multiple icons in them. For instance here is the icon browse for the filezilla.exe which was the first shortcut my mouse found:

filezilla change icon

The FZ icon select will be index 0 and the other icons will increment from there.

btm: Ah, cool. That reference document wasn't has helpful as you were. Could you update the description something like this?

```
- iconlocation: Icon to use, in the format of "path, index" where index is which icon in that file to use (See [WshShortcut.IconLocation](https://msdn.microsoft.com/en-us/library/3s9bx7at.aspx))
```

lambacck: Done.

PR #180, opened 2015-02-16
Escaping fix for README.md

docwhat:
The `windows_task` examples didn't render right because of the backslashes (`\`). I just fixed them.

Trivial change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -522,8 +522,8 @@ Run Chef every 15 minutes
 windows_task 'Chef client' do
   user 'Administrator'
   password '$ecR3t'
-  cwd 'C:\chef\bin'
-  command 'chef-client -L C:\tmp\'
+  cwd 'C:\\chef\\bin'
+  command 'chef-client -L C:\\tmp\\'
   run_level :highest
   frequency :minute
   frequency_modifier 15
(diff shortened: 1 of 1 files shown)


btm:
By render, do you mean in the way that Github displays the markdown, or how Chef interprets them? A single backslash should be fine when it is inside single quotes, as Ruby doesn't consider backslash to be an escape character in single quotes:

```
irb(main):001:0> 'C:\foo\bar'
=> "C:\\foo\\bar"
```

Given a recipe of:
```
directory 'C:\test'
```

Chef runs correctly:
```
> chef-apply foo.rb
Recipe: (chef-apply cookbook)::(chef-apply recipe)
  * directory[C:\test] action create
    - create new directory C:\test
```

jdmundrawala:
@btm the problem is that there are strings ending with `\`, so the interpreter thinks you want to escape the `'`

jdmundrawala:
basically, only the `command` attributes need fixing in this example, and they need to be fixed by escaping the `\`. For example:
```ruby
windows_task 'Chef client' do
  user 'Administrator'
  password '$ecR3t'
  cwd 'C:\chef\bin'
  command 'chef-client -L C:\tmp\\'
  run_level :highest
  frequency :minute
  frequency_modifier 15
end
```

docwhat:
@jdmundrawala 

If you put `\\` inside a single quote, it is interpreted as `\`.  Example:

```
irb> puts 'c:\\foo\\'
c:\foo\
nil
```

So the change I made is also legit ruby.

PR #190, opened 2015-03-04
Create user home resource

dhoer:
Creates and populates home directory without requiring logon.  Inspiration taken from https://github.com/opscode-cookbooks/omnibus/blob/master/recipes/_user.rb#L48-62.

### DIFF
--- a/providers/home.rb
+++ b/providers/home.rb
@@ -2,26 +2,38 @@ require 'chef/mixin/shell_out'
 include Chef::Mixin::ShellOut
 
 action :create do
-  task_name = "create_#{@new_resource.name}_home"
-  user_home = ::File.join(ENV['SYSTEMDRIVE'], 'Users', @new_resource.name)
-    .gsub(::File::SEPARATOR, ::File::ALT_SEPARATOR)
+  task_name = "build_#{new_resource.name}_home"
+  user_home = ::File.join(ENV['SYSTEMDRIVE'], 'Users', new_resource.name)
+              .gsub(::File::SEPARATOR, ::File::ALT_SEPARATOR)
 
-  # Run whoami as scheduled task to get LocalServer group privilege:
+  # Create whoami as scheduled task to get Local Server group privilege:
   # SeAssignPrimaryTokenPrivilege (Replace a process-level token)
-  if task_query(task_name).empty?
-    cmd =  "schtasks /Create /TN \"#{task_name}\" "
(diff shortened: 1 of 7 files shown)


dhoer:
Is there interest in adding this feature to windows cookbook, or should does it belong somewhere else?

jdmundrawala:
Looks interesting, I think it would be a nice addition

dhoer:
Awesome! When do you think it would be released?  I have it as a dependency in windows_desktop::headless_display cookbook: https://github.com/dhoer/chef-windows_desktop/blob/master/recipes/headless_display.rb#L27-30

providers/home.rb
```diff
@@ -0,0 +1,46 @@
...
+  task_name = "create_#{@new_resource.name}_home"
+  user_home = ::File.join(ENV['SYSTEMDRIVE'], 'Users', @new_resource.name)
+    .gsub(::File::SEPARATOR, ::File::ALT_SEPARATOR)
+
+  # Run whoami as scheduled task to get LocalServer group privilege:
+  # SeAssignPrimaryTokenPrivilege (Replace a process-level token)
+  if task_query(task_name).empty?
+    cmd =  "schtasks /Create /TN \"#{task_name}\" "
+    cmd += "/SC once /SD \"01/01/2003\" /ST \"00:00\" /TR \"whoami.exe\" "
+    cmd += "/RU \"#{@new_resource.name}\" /RP \"#{@new_resource.password}\" "
```
jdmundrawala: won't this password be available for all to see?

dhoer: You are correct. Good catch.  I can delete this task after it completes.  Or is there a better way to run whoami with SeAssignPrimaryTokenPrivilege?

dhoer: Wait, I don't know if this password is visible or not. I can't tell from the documentation: https://msdn.microsoft.com/en-us/library/windows/desktop/bb736357%28v=vs.85%29.aspx.  It still would make sense to delete the task to be on the safe side.

dhoer: I can try to run execute resource with sensitive turned on, since I don't see that option for shell_out.

jdmundrawala: As far as I know, all that does is prevent chef from reporting the sensitive information. I'm not really sure what the right thing to do here is. I'll ping people, see what they think

dhoer:
@jdmundrawala What do you think of this solution?

dhoer:
@jdmundrawala  Thanks for the code review and making this better.  I'm closing this pull request.  You can still add it in the future, if you want.  This home feature is now available in windows_desktop cookbook.

PR #194, opened 2015-03-26
Update testing stack with the latest test-kitchen changes.

cmluciano:
A new parameter was added to the driver config so that the correct
communicator is setup. Technically transport is not needed anymore
since it is discovered with the name, but I left it in there to be
explicit.

I bumped the gems to the latest working betas. Kitchen-vagrant
0.17.0.beta3 requires the vagrant-winrm plugin and will display
a message stating that you need to install it if you have not
already. I think this should suffice as a warning, but I could add
a special testing section to the README if necessary.

This has been tested with all 3 suites.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,8 +1,8 @@
-source "https://rubygems.org"
+source 'https://rubygems.org'
 
 group :development do
-  gem "test-kitchen", :git => 'https://github.com/test-kitchen/test-kitchen.git', :branch => 'windows-guest-support'
-  gem 'kitchen-vagrant', git: 'https://github.com/test-kitchen/kitchen-vagrant.git', :branch => 'windows-guest-support'
-  gem "berkshelf"
-  gem "vagrant-wrapper", ">= 2.0"
+  gem 'test-kitchen',     '1.4.0.beta.2'
+  gem 'kitchen-vagrant',  '0.17.0.beta.3'
+  gem 'berkshelf'
+  gem 'vagrant-wrapper',  '>= 2.0'
 end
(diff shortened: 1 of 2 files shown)


cmluciano:
cc @ksubrama @smurawski

Gemfile
```diff
@@ -1,8 +1,8 @@
-source "https://rubygems.org"
+source 'https://rubygems.org'
 
 group :development do
-  gem "test-kitchen", :git => 'https://github.com/test-kitchen/test-kitchen.git', :branch => 'windows-guest-support'
-  gem 'kitchen-vagrant', git: 'https://github.com/test-kitchen/kitchen-vagrant.git', :branch => 'windows-guest-support'
-  gem "berkshelf"
-  gem "vagrant-wrapper", ">= 2.0"
+  gem 'test-kitchen',     '1.4.0.beta.2'
+  gem 'kitchen-vagrant',  '0.17.0.beta.3'
```
btm: can we specify something "like >= 0.17.0.beta.3" so we don't have to change this file if we do another beta?

cmluciano: Indeed, @fnichol just released beta.4. Testing now and then will push up with ">="

fnichol:
@cmluciano Nice! Seeing as you're in there anyway, think you could drop the `forward_agent` line as well? I don't know where that came from but doesn't really relate to anything currently

cmluciano:
@fnichol will do

cmluciano:
Should be good to go with @btm and @fnichol changes

ksubrama:
Rebased, cleaned up and merged in c1acb618c6e2eb8987f506ed5f0b03a8aae9c837.

PR #197, opened 2015-05-01
enable windows_task to accept pathed task names + update test-kitchen

trevorghess:
This allows windows_task to accept the following task name formats:
```ruby
"Task Name"
"/Task Name"
"Path/Task Name"
"/Path/Task Name"
```

It also updates the test-kitchen references to work with 1.4.0 so that the tests can run

### DIFF
--- a/test/cookbooks/minimal/recipes/default.rb
+++ b/test/cookbooks/minimal/recipes/default.rb
@@ -29,3 +29,28 @@ windows_task 'disable chef test' do
   name 'chef test'
   action :disable
 end
+
+windows_task 'create chef test' do
+  name 'chef\\chef test'
+  action :create
+  cwd "C:\\"
+  command 'dir'
+end
+
+windows_task 'disable chef test' do
+  name 'chef\\chef test'
+  action :disable
+end
+
+windows_task 'create chef test' do
+  name '\\chef\\chef test2'
+  action :create
+  cwd "C:\\"
+  command 'dir'
+end
+
+windows_task 'disable chef test' do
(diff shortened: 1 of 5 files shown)


test/cookbooks/minimal/recipes/default.rb
```diff
@@ -29,3 +29,28 @@
...
+  command 'dir'
+end
+
+windows_task 'disable chef test' do
+  name 'chef\\chef test'
+  action :disable
+end
+
+windows_task 'create chef test' do
+  name '\\chef\\chef test2'
```
smurawski: Why do we need to escape slashes when using a single quoted string?  This isn't intuitive nor how the native commands represent the path of the scheduled tasks.

test/cookbooks/minimal/recipes/default.rb
```diff
@@ -29,3 +29,28 @@
   name 'chef test'
   action :disable
 end
+
+windows_task 'create chef test' do
+  name 'chef\\chef test'
+  action :create
+  cwd "C:\\"
```
smurawski: Did you verify that cwd was actually being set?  I'm not seeing that working in my tests.

smurawski:
@trevorghess I'm in the process of getting this merged in (see #204) along with some more tests.  I have a few more items to add and we'll cut a new release of this cookbook.

smurawski:
Merged with #204

smurawski:
I'll cut a new release of this cookbook on 5/14/2015.

PR #199, opened 2015-05-08
use Chef.set_resource_priority_array to suppress warning

lamont-granquist:


jdmundrawala:
do we need to set the provider priority as well?

lamont-granquist:
@someara @jdmundrawala @smurawski could use some eyes on this.

see chef/chef#3349 for why undefining the existing class is bad.

lamont-granquist:
Does it wire itself up explicitly?

jdmundrawala:
Looks like it:
https://github.com/opscode-cookbooks/windows/blob/master/libraries/windows_package.rb#L215

lamont-granquist:
The cookbook version hardcodes itself to the cookbook version of the provider, yeah

jdmundrawala:
:+1:

lamont-granquist:
If people want to do both they're gonna have to setup a provides alias anyway to wire up a way to hit the internal core windows_package resource, so I think there's nothing we need to do to the provider mapping there either.

smurawski:
:+1:

jdmundrawala:
I don't think people have ever done that

lamont-granquist:
The guy in chef/chef#3349 wants to use both of them...

PR #200, opened 2015-05-11
Fix resource priority

jdmundrawala:
I think the old logic would never match?

cc @lamont-granquist

### DIFF
--- a/libraries/windows_package.rb
+++ b/libraries/windows_package.rb
@@ -219,17 +219,17 @@ class Chef
 end
 
 if Gem::Version.new(Chef::VERSION) < Gem::Version.new('12')
+  # this wires up the cookbook version of the windows_package resource as Chef::Resource::WindowsPackage,
+  # which is kinda hella janky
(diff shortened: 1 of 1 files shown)


lamont-granquist:
No the `Chef.set_resource_priority_array` is `chef >= 12.3.0` so I'm trying to catch `12.0.0 <= chef_version < 12.3.0` with the else block.

lamont-granquist:
(we don't have dynamic *resource* resolution with the priority array stuff before 12.3.0, and don't have the chef_class stuff before then either)

jdmundrawala:
Right, but it looks like `Chef.respond_to?(:set_resource_priority_array)` check only happens
if `Gem::Version.new(Chef::VERSION) < Gem::Version.new('12')`.

```ruby
if Gem::Version.new(Chef::VERSION) < Gem::Version.new('12')
  if Chef.respond_to?(:set_resource_priority_array)
   ....
```
Since `Gem::Version.new('12.3.0') < Gem::Version.new('12') == false`, `if Chef.respond_to?(:set_resource_priority_array)` will never run.

lamont-granquist:
lol, yes, i can't read.

:+1:

PR #203, opened 2015-05-12
Move testing to the Pester Verifier

smurawski:
Move existing tests to Pester and change the verifier from Busser to Pester.
Update Gemfile after release of tk with guest support
Rename default suite to minimal (for the recipe tested)


jdmundrawala:
:+1:

Issue #111, opened 2014-09-03
commands on path are not available after running windows_path


dawilliams:
Chef:                           11.14.6
Windows Cookbook:   1.34.2
Node :                         Windows 2012
I’ve encountered a scenario where after the windows_path resource is ran in a Chef run, other resources can't find items currently on the Windows path.  For example the batch resource fails with “STDERR: ‘cmd.exe’ is not recognized as an internal or external command.”  If you don’t have the windows_path resource in your recipe or if you run the recipe again, the batch resource can find the cmd command on the path.

carpnick:
+1

Using 1.36.1.  The path variable is being updated, but a reboot is required to activate all the changes.  Feels like an API call is being missed to notify all other processes like explorer.  Hitting this when installing git. and java

jdmundrawala:
@carpnick what version of chef?

carpnick:
@jdmundrawala  chef-client 11.16.4

jdmundrawala:
@carpnick 
https://github.com/chef/chef/blob/master/lib/chef/provider/env/windows.rb#L39
It looks like it should do the right thing.

Can you send me a simple recipe that fails and its cookbook dependencies/versions

carpnick:
@jdmundrawala 
Sure want me to just create some gists or do you have another delivery in mind?

btm:
@carpnick a gist is fine. if it's short, just dump it in here and surround it with \`\`\` blocks

carpnick:
*Installs Java and Git
Gist for recipe: https://gist.github.com/carpnick/0da3f884e4825ac6af64

metadata.rb:
depends           'jenkins', 	'~> 2.2' 
depends           'git', 	'~> 4.1' 
depends           'java', 	'~> 1.29' 

Actual in use versions:
https://gist.github.com/carpnick/0572bccf90269a3271bd

sneal:
I just switched from using setx on the command line via an execute resource to using the windows_path resource and noticed the same behavior. Setx will make the env var change available to all new processes without requiring a reboot (or perhaps just a re-login), while windows_path requires a reboot. 

So there's a way to do it on Windows. Perhaps I'll fire up Spy++ and see what setx is doing that's different.

sneal:
It seems the way Chef handles notifying the system of an env var change works correctly. This code example has the same behavior as setx (for system vars), meaning the new env var is immediately available to new processes without needing to relogon.

```ruby
[...]

PR #208, opened 2015-05-14
Add some tests for windows_path

smurawski:
Start a baseline of tests to validate the discussion around #111 (I've tested with 11.16.4 and 12.3.0)

/cc @pburkholder @sneal @btm @jdmundrawala @ksubrama @carpnick @spudnic @dawilliams @whoisjake


jdmundrawala:
yes please

lamont-granquist:
I think your integration test is only testing that chef-client itself picked up the updated path.  That does not check for the case in #111.  You need to setup a system path to a directory which has a runnable file in it.  Then in your integration test actually try to run that file and see if it succeeds.

smurawski:
@lamont-granquist That was a poorly named test.  It's actually checking the registry where new processes get their environmental variables from, not just child processes.

test/integration/path/pester/minimal.path.Tests.ps1
```diff
@@ -0,0 +1,25 @@
...
+      ($Paths -contains 'C:\path_test_path') | should be $true
+    }
+
+    it "'C:\path_test_another_path' was added to the path" {
+      ($Paths -contains 'C:\path_test_another_path') | should be $true
+    }
+
+    it 'Child processes and shellouts have an updated path' {
+      'c:\paths.txt' | should contain 'C:\\path_test_another_path'
+      'c:\paths.txt' | should contain 'C:\\path_test_path'
```
lamont-granquist: stray `\\`'s here left over from old double quotes?

lamont-granquist:
Yeah, but the registry is getting updated, and check on paths.txt is effectively just checking that chef-client updates its own path.  It doesn't check that an external powershell window really picks up the registry setting.  You should have the chef recipe drop a hello world script somewhere and then setup the windows_path to its directory and then have the powershell script try to invoke it through the path.

test/integration/path/pester/minimal.path.Tests.ps1
smurawski: the contain matcher is a regex.

smurawski:
Ok, there are now tests that validate the updated path is being seen by newly launched external programs.  I use task scheduler to launch a new PowerShell session and log the path.

This fails on Server 2008 R2, but succeeds on Server 2012 and 2012 R2.

ksubrama:
:+1:

PR #209, opened 2015-05-15
DNS LWRP

rlaveycal:
A new LWRP that manages A and CNAME records in Windows DNS using the DNSCMD command.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,6 +16,7 @@ Version 1.3.0+ of this cookbook requires Chef 0.10.10+.
 * Windows Server 2008 (R1, R2)
 
 The `windows_task` LWRP requires Windows Server 2008 due to its API usage.
+The `windows_dns` LWRP requires DNSCMD.exe to be installed.
 
 ### Cookbooks
 The following cookbooks provided by Chef Software are required as noted:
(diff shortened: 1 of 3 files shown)


providers/dns.rb
```diff
@@ -0,0 +1,154 @@
...
+include Chef::Mixin::ShellOut
+include Windows::Helper
+
+# Support whyrun
+def whyrun_supported?
+  true
+end
+
+action :create do
+	if @current_resource.exists
```
ksubrama: Could you change \t to two spaces to be consistent with everything else in the cookbook?

providers/dns.rb
```diff
@@ -0,0 +1,154 @@
...
+	parts = @current_resource.host_name.scan(/(\w+)\.(.*)/)
+	@host = parts[0][0]
+	@domain = parts[0][1]
+
+	getAttributes
+end
+
+private
+def getAttributes
+	@command = locate_sysnative_cmd("dnscmd.exe")
```
ksubrama: Isn't dnscmd deprecated?  I don't have much experience in the area but there seem to be equivalent powershell commands like Get-DnsServerResourceRecord that provide typed data and lets us avoid parsing command output.

README.md
```diff
@@ -90,6 +91,47 @@ windows_batch 'echo some env vars' do
 end
 ```
 
+### windows_dns
+Configures A and CNAME records in Windows DNS. This requires the DNSCMD to be installed, which is done by adding the DNS role to the server or installing the Remote Server Admin Tools.
+
+#### Actions
+- :create: creates/updates the DNS entry
```
ksubrama: A :install action that simply calls windows_feature to install the dns server might solve the requirement that DNSCMD.exe be installed.

ksubrama:
Thanks a lot for contributing this code!

/cc @smurawski thoughts on the powershell vs dnscmd approach?

providers/dns.rb
rlaveycal: Get-DnsServerResourceRecord seems to be Windows 2012 only...I need to target 2008R2

tas50:
We should have merged this a while ago. Thanks for the contribution

Issue #211, opened 2015-05-19
Allow certificates to be maintained and bound to HTTP ports


rlaveycal:
We should be able to import certificate files (cer, pfx, etc.), grant access to private keys, and bind them to ports to allow TLS to function.

#167 is one such implementation...another will follow shortly.

PR #212, opened 2015-05-19
support installing certificates and binding to ports

rlaveycal:
Implements #211 

Allows
- Cert to be imported from cer, p7b and pfx files
- Add cert to personal, intermediate and root stores
- Add cert to machine or user stores
- Grant access to private key
- Bind a certificate to a port to enable TLS

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,15 +95,15 @@ end
 Installs a certificate into the Windows certificate store from a file, and grants read-only access to the private key for designated accounts.
 
 #### Actions
-- :create: creates or updates a certificate
-- :delete: deletes a certificate
-- :acl_add: adds read-only entries to a certificte's private key ACL
+- :create: creates or updates a certificate.
+- :delete: deletes a certificate.
+- :acl_add: adds read-only entries to a certificate's private key ACL.
 
 #### Attribute Parameters
 - source: name attribute. The source file (for create and acl_add), thumprint (for delete and acl_add) or subject (for delete).
 - pfx_password: the password to access the source if it is a pfx file.
-- private_key_acl: array of 'domain\account' entries to be granted read-only access to the certificte's private key. This is not idempotent.
-- store_name: the certificate store to maniplate. One of MY (default : personal store), CA (trusted intermediate store) or ROOT (trusted root store)
+- private_key_acl: array of 'domain\account' entries to be granted read-only access to the certificate's private key. This is not idempotent.
+- store_name: the certificate store to maniplate. One of MY (default : personal store), CA (trusted intermediate store) or ROOT (trusted root store).
 - user_store: if false (default) then use the local machine store; if true then use the current user's store.
 
 #### Examples
@@ -134,16 +134,16 @@ end
 Binds a certificate to an HTTP port in order to enable TLS communication.
 
 #### Actions
-- :create: creates or updates a binding
-- :delete: deletes a binding
+- :create: creates or updates a binding.
+- :delete: deletes a binding.
 
 #### Attribute Parameters
 - cert_name: name attribute. The thumprint(hash) or subject that identifies the certicate to be bound.
-- name_kind: indicates the type of cert_name. One of :subject (default) or :hash
-- address: the address to bind against. Default is 0.0.0.0 (all IP addresses)
+- name_kind: indicates the type of cert_name. One of :subject (default) or :hash.
+- address: the address to bind against. Default is 0.0.0.0 (all IP addresses).
 - port: the port to bind against. Default is 443.
 - app_id: the GUID that defines the application that owns the binding. Default is the values used by IIS.
-- store_name: the store to locate the certificate in. One of MY (default : personal store), CA (trusted intermediate store) or ROOT (trusted root store)
+- store_name: the store to locate the certificate in. One of MY (default : personal store), CA (trusted intermediate store) or ROOT (trusted root store).
 
 #### Examples
 ```ruby
--- a/README.md
+++ b/README.md
@@ -90,6 +90,78 @@ windows_batch 'echo some env vars' do
 end
 ```
 
+### windows_certificate
+
+Installs a certificate into the Windows certificate store from a file, and grants read-only access to the private key for designated accounts
+
+#### Actions
+- :create: creates or updates a certificate
+- :delete: deletes a certificate
+- :acl_add: adds read-only entries to a certificte's private key ACL
+
+#### Attribute Parameters
+- source: name attribute. The source file (for create and acl_add) or thumprint/subject (for delete and acl_add).
+- pfx_password: the password to access the source if it is a pfx file.
+- private_key_acl: array of 'domain\account' entries to be granted read-only access to the certificte's private key. This is not idempotent.
+- store_name: the certificate store to maniplate. One of MY (default : personal store), CA (trusted intermediate store) or ROOT (trusted root store)
+- user_store: if false (default) then use the local machine store; if true then use the current user's store.
+
+#### Examples
+```ruby
+# Add PFX cert to local machine personal store and grant accounts read-only access to private key
+windows_certificate "c:/test/mycert.pfx" do
+	pfx_password	"password"
+	private_key_acl	["acme\fred", "pc\jane"]
+end
+```
+
+```ruby
+# Add cert to trusted intermediate store
+windows_certificate "c:/test/mycert.cer" do
+	store_name	"CA"
+end
+```
+
+```ruby
+# Remove all certicates matching the subject
+windows_certificate "me.acme.com" do
+	action :delete
+end
+```
+
+### windows_certificate_binding
+
+Binds a certificate to an HTTP port in order to enable TLS communication.
+
+#### Actions
+- :create: creates or updates a binding
+- :delete: deletes a binding
+
+#### Attribute Parameters
+- cert_name: name attribute. The thumprint(hash) or subject that identifies the certicate to be bound.
+- name_kind: indicates the type of cert_name. One of :subject (default) or :hash
+- address: the address to bind against. Default is 0.0.0.0 (all IP addresses)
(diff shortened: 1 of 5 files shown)


providers/certificate.rb
```diff
@@ -0,0 +1,160 @@
...
+	if file.downcase.end_with?('pfx') && (@new_resource.pfx_password.nil? || @new_resource.pfx_password.empty?)
+		raise "No password given for PFX file"
+	end
+	
+	password = " -p #{@new_resource.pfx_password}" if file.downcase.end_with?('pfx')
+	cmd = shell_out("#{@command}#{password} \"#{file}\"")
+	Chef::Log.debug "certutil reports: #{cmd.stdout}"
+	
+	# extract values from returned text
+	if cmd.exitstatus == 0
```
ksubrama: If you're already using powershell features below, you can make some of this cleaner by using (New-Object System.Security.Cryptography.X509Certificates.X509Certificate "file").GetCertHashString() in a powershell_script.  Actually, now that I think of it, a lot of this can be simplified by using powershell_script and putting the checks in guard clauses.

providers/certificate.rb
```diff
@@ -0,0 +1,160 @@
...
+end
+
+def updateAcl!(certHash)
+	if (@new_resource.private_key_acl.nil? || @new_resource.private_key_acl.length == 0)
+		return
+	end
+	
+	raise "ACL can only be set on local machine certificates" if @new_resource.user_store
+	
+	# this PS came from http://blogs.technet.com/b/operationsguy/archive/2010/11/29/provide-access-to-private-keys-commandline-vs-powershell.aspx
```
ksubrama: Put this in a powershell_script code block so that you don't have to append things to a string,

providers/certificate.rb
```diff
@@ -0,0 +1,160 @@
...
+			raise "Couldn't find hash in output : #{cmd.stdout}"
+		end
+	else
+		raise "certutil returned error #{cmd.exitstatus} : #{cmd.stderr} #{cmd.stdout}"
+	end
+end
+
+def isCertInStore?(certHash)
+	user = " -user" if @new_resource.user_store
+	cmd = shell_out("#{@command}#{user} -store #{@new_resource.store_name} \"#{certHash}\"")
```
ksubrama: This can become "Test-Path Cert:\CurrentUser\{new_resource.store_name}\{certHash}" in a guard clause.  Or LocalMachine instead of CurrentUser based on user_store.

README.md
```diff
@@ -90,6 +90,78 @@ windows_batch 'echo some env vars' do
...
+```ruby
+# Remove all certicates matching the subject
+windows_certificate "me.acme.com" do
+	action :delete
+end
+```
+
+### windows_certificate_binding
+
+Binds a certificate to an HTTP port in order to enable TLS communication.
```
ksubrama: Hmm - wouldn't this be more in place within the IIS or windows_firewall cookbooks?

ksubrama:
Thank you so much for the contribution.  /cc @smurawski for your thoughts as well.

rlaveycal:
Thanks for the feedback. I've re-written to use inner powershell_script resources. Definitely neater not parsing command output. Is there a way to set updated_by_last_action to the powershell_script's value?

ksubrama:
Yes there is!  https://docs.chef.io/lwrp.html#use-inline-resources
This will automatically delegate the responsibility to notify/fire-updates to the underlying resources.

rlaveycal:
Sweet :)

providers/certificate.rb
```diff
@@ -0,0 +1,142 @@
...
+  powershell_script @new_resource.name do
+    code codeScript
+    only_if onlyifScript
+  end
+end
+
+action :delete do
+  # We can do everything in a powershell script resource
+  subject = @current_resource.source.sub(/\*/, '`*') # escape any * in the source
+  certCommand = "gci cert:\\#{@location}\\#{@new_resource.store_name} | where { $_.Subject -like '*#{subject}*' -or $_.Thumbprint -eq '#{subject}'}"
```
ksubrama: Thumbprint -eq @current_resource.source

providers/certificate.rb
```diff
@@ -0,0 +1,142 @@
...
+    not_if notifScript
+  end
+end
+
+# acl_add is a modify-if-exists operation : not idempotent
+action :acl_add do
+  # We can do everything in a powershell script resource
+  hash = nil
+
+  if ::File.exists?(@new_resource.source)
```
ksubrama: This file might not exist yet if it's going to be created by an earlier resource doing a :create.  I'd recommend checking if source is a hash (like the regex below), and basing your script on that.

providers/certificate.rb
```diff
@@ -0,0 +1,142 @@
...
+
+private
+def getAclScript(hash)
+  if (!@new_resource.private_key_acl.nil? && @new_resource.private_key_acl.length > 0)
+    raise "ACL can only be set on local machine certificates" if @new_resource.user_store
+
+    # this PS came from http://blogs.technet.com/b/operationsguy/archive/2010/11/29/provide-access-to-private-keys-commandline-vs-powershell.aspx
+    setAclScript = <<-EOH
+      $hash = #{hash}
+      $storeCert = gci "cert:\\#{@location}\\#{@new_resource.store_name}\\$hash"
```
ksubrama: $hash isn't used again - so you can just substitute that variable here.

providers/certificate.rb
```diff
@@ -0,0 +1,142 @@
...
+      if ($storeCert -eq $null) { throw 'no key exists.' } 
+      $keyname = $storeCert.PrivateKey.CspKeyContainerInfo.UniqueKeyContainerName
+      if ($keyname -eq $null) { throw 'no private key exists.' } 
+      $fullpath = "$env:ProgramData\\Microsoft\\Crypto\\RSA\\MachineKeys\\$keyname"
+    EOH
+    @new_resource.private_key_acl.each do | name |
+      setAclScript << "$uname='#{name}'; icacls $fullpath /grant $uname`:RX;"
```
ksubrama: Oooh...  I think I understand what you're attempting to do.  You're just granting read access to the underlying stored key and this is just a way to find out the real path.  We have idempotent acl support for windows in the file resource.  See https://docs.chef.io/resource_file.html#windows-file-security

Now that I understand that, there's a slightly better way here.  You still need to run that powershell script to get the full path.  For that you would want to use PowershellOut.  But after you get the fullPath (from stdout of that command), you declare a file resource with the correct path and just pass "rights :read_execute, private_key_acl" to give that array to the resource and let Chef handle it.

If this seems like too big a change, you can leave it as is and I can come back to it during a clean up pass.

ksubrama:
Thank you for the high quality patch!

providers/certificate_binding.rb
```diff
@@ -0,0 +1,121 @@
...
+  getCurrentHash
+end
+
+private
+def getCurrentHash()
+  cmd = shell_out("#{@command} http show sslcert ipport=#{@current_resource.address}:#{@current_resource.port}")
+  Chef::Log.debug "netsh reports: #{cmd.stdout}"
+
+  if cmd.exitstatus == 0
+    m = cmd.stdout.scan(/Certificate Hash\s+:\s?(\w{40})/)
```
ksubrama: is [A-Fa-f0-9] to ensure it's hex?

README.md
```diff
@@ -90,6 +90,78 @@ windows_batch 'echo some env vars' do
...
 ```
 
+### windows_certificate
+
+Installs a certificate into the Windows certificate store from a file, and grants read-only access to the private key for designated accounts.
+
+#### Actions
+- :create: creates or updates a certificate
+- :delete: deletes a certificate
+- :acl_add: adds read-only entries to a certificte's private key ACL
```
ksubrama: typo: certificate

README.md
```diff
@@ -90,6 +90,78 @@ windows_batch 'echo some env vars' do
 end
 ```
 
+### windows_certificate
+
+Installs a certificate into the Windows certificate store from a file, and grants read-only access to the private key for designated accounts.
+
+#### Actions
+- :create: creates or updates a certificate
```
ksubrama: nit: Period at the end of sentences.

README.md
```diff
@@ -90,6 +90,78 @@ windows_batch 'echo some env vars' do
...
+- :delete: deletes a certificate
+- :acl_add: adds read-only entries to a certificte's private key ACL
+
+#### Attribute Parameters
+- source: name attribute. The source file (for create and acl_add), thumprint (for delete and acl_add) or subject (for delete).
+- pfx_password: the password to access the source if it is a pfx file.
+- private_key_acl: array of 'domain\account' entries to be granted read-only access to the certificte's private key. This is not idempotent.
```
ksubrama: typo: certificate

providers/certificate.rb
rlaveycal: Yes. A complication is that the private key ACL has read/write permissions granted to admin users by default which need to be left in place. That means either that the user has to remember to include those in the acl list or we need an idempotent function on just those acl members with read permission. Probably best left for further thought.

providers/certificate.rb
rlaveycal: For create the source can only be a file...if it doesn't exist then you get an exception from PS ```STDERR: Exception calling ".ctor" with "1" argument(s): "The system cannot find the path specified."```. It works fine using a remote_file to create the file.

providers/certificate.rb
rlaveycal: true, but ```hash``` is either a simple declaration or an expression (line 68) so for readability/sanity I declared $hash to force evaluation.

providers/certificate.rb
ksubrama: Ah - the rabbit hole goes deeper.  This seems to be the best approach for now.

ksubrama:
Thank you so much!

Issue #213, opened 2015-05-19
Allow HTTP ACLs to be maintained


rlaveycal:
The netsh command's urlacl option allows non-admin users to be granted the rights to open HTTP endpoints. We should be able to define these permissions through Chef

PR #214, opened 2015-05-19
Implement LWRP to maintain HTTP ACLs

rlaveycal:
Implements #213

### DIFF
--- a/README.md
+++ b/README.md
@@ -173,6 +173,31 @@ Font files should be included in the cookbooks
 windows_font 'Code New Roman.otf'
 ```
 
+### windows_http_acl
+Sets the Access Control List for an http URL to grant non-admin accounts permission to open HTTP endpoints
+
+#### Actions
+- :create: creates or updates the ACL for a URL
+- :delete: deletes the ACL from a URL.
+
+#### Attribute Parameters
+- url: the name of the url to be created/deleted.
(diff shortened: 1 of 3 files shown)


ksubrama:
Thanks for this pull request!

I know that WCF is technically different from the Windows firewall and from IIS, but I think this would fit in better as a resource in https://github.com/lynx44/windows_firewall or a similar cookbook.  Perhaps even https://github.com/opscode-cookbooks/iis ?

README.md
```diff
@@ -173,6 +173,31 @@ Font files should be included in the cookbooks
 windows_font 'Code New Roman.otf'
 ```
 
+### windows_http_acl
+Sets the Access Control List for an http URL to grant non-admin accounts permission to open HTTP endpoints
+
+#### Actions
+- :create: creates or updates the ACL for a URL
```
ksubrama: Can you leave a comment mentioning that acls granted to other users will be removed?

ksubrama:
Alright - let's put it in for now.  If it grows more features, we can extract it to a cookbook on its own.  Other than that one comment change, this looks good!

README.md
```diff
@@ -173,6 +173,31 @@ Font files should be included in the cookbooks
 windows_font 'Code New Roman.otf'
 ```
 
+### windows_http_acl
+Sets the Access Control List for an http URL to grant non-admin accounts permission to open HTTP endpoints
```
ksubrama: nit: period at the end of sentences.

PR #218, opened 2015-05-26
Fix #139 by giving windows_reboot better defaults.

### LINKED ISSUE #139: windows_reboot resource does not execute if timeout is not specified and the name of the resource is a string
This does not result in a reboot.

windows_reboot "anything" do
    reason 'Reboot'
    action :request
    only_if { reboot_pending? }
end

This does result in a reboot

windows_reboot "something" do
    timeout 1
    reason 'Reboot'
    action :request
    only_if { reboot_pending? }
end

ksubrama:


resources/reboot.rb
```diff
@@ -19,11 +19,7 @@
...
+default_action :request
 
 attribute :timeout, :kind_of => Integer, :name_attribute => true
-attribute :reason, :kind_of => String, :default => ''
-
-def initialize(name,run_context=nil)
-  super
-  @action = :request
-end
+attribute :reason, :kind_of => String, :default => 'Chef Software Chef initiated reboot'
```
jdmundrawala: Chef Client run

jdmundrawala:
:+1:

PR #220, opened 2015-05-27
Fix #219

### LINKED ISSUE #219: Windows 1.37.0 with PR#204 is breaking existing cookbooks using windows_task resource
Unfortunately, Pull Request#204 breaks all of our existing internal cookbooks with the windows_task resource using this new pr as well as chef-client::task...

https://github.com/opscode-cookbooks/chef-client/issues/313

jdmundrawala:
The problem is we try to set password on new_resource to nil.
The correct way to change a value is to use (), and not =. However,
you can't use (nil) because that means return me the current value.

### DIFF
--- a/providers/task.rb
+++ b/providers/task.rb
@@ -41,7 +41,7 @@ action :create do
     options['ST'] = @new_resource.start_time unless @new_resource.start_time.nil?
     options['TR'] = "\"#{@new_resource.command}\" "
     options['RU'] = @new_resource.user
-    options['RP'] = @new_resource.password if @new_resource.password
+    options['RP'] = @new_resource.password if use_password?
     options['RL'] = 'HIGHEST' if @new_resource.run_level == :highest
     options['IT'] = '' if @new_resource.interactive_enabled
     options['D'] = @new_resource.day if @new_resource.day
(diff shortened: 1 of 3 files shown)


providers/task.rb
```diff
@@ -241,3 +240,7 @@ def validate_create_day
     end
   end
 end
+
+def password
+  @new_resource.password if !SYSTEM_USERS.include?(@new_resource.user.upcase)
```
btm: unless instead of "if !"

providers/task.rb
```diff
@@ -241,3 +240,7 @@ def validate_create_day
     end
   end
 end
+
+def password
```
btm: `#password_unless_user_exists` ?

providers/task.rb
```diff
@@ -41,7 +41,7 @@
     options['ST'] = @new_resource.start_time unless @new_resource.start_time.nil?
     options['TR'] = "\"#{@new_resource.command}\" "
     options['RU'] = @new_resource.user
-    options['RP'] = @new_resource.password if @new_resource.password
+    options['RP'] = password if password
```
btm: "password if password" sounds like we're setting the password if we have one. it'd be nice to change the name of the method to something that conveys more of its logic.

ksubrama:
:+1: modulo btm's comments.

jdmundrawala:
@btm better

btm:
:+1:

PR #221, opened 2015-06-02
Certificate delete doesn't run on PS 2

rlaveycal:
Powershell 2 doesn't treat a single item as an array meaning count returns null. Need to coerce to an array.


ksubrama:
Thank you!

PR #229, opened 2015-06-10
Fix certificate provider failing when being used remotely

### LINKED ISSUE #227: windows_certificate remotely over winrm does not work
As the title says, using the windows_certificate resource over remotely over winrm doesn't work. Running locally on the machine succeeds. Remote execution fails with "Begin output of "powershell.exe" -NoLogo -NonInteractive -NoProfile -ExecutionPolicy Unrestricted -InputFormat None -File "C:/Users/ADMINI~1/AppData/Local/Temp/chef-script20150609-2888-i668tc.ps1"
STDERR: Exception calling ".ctor" with "3" argument(s): "Access denied."

Full output: http://pastebin.com/R3T4wkSb

I copied the script to the host and ran it remotely, with the same error: http://pastebin.com/cjsGXGW0

Now, this is where things get weird. If I run that exact script locally (on the remote machine), not only does  it work,  but then it also starts working remotely. Even if after removing the cert, both remote and local continue function and chef runs complete without error. The only way I can reproduce this is with a fresh host. I had my own implementation of this resource working fine. So its something specific to how this is written.

Note: this is not the first time I've run across windows treating the same code very differently local vs remote: https://github.com/opscode-cookbooks/windows/issues/196

tophercullen:
Using the certificate resource remotely, like winrm, would fail
while the exact same code would run properly locally.
Fixes #227

### DIFF
--- a/providers/certificate.rb
+++ b/providers/certificate.rb
@@ -35,7 +35,7 @@ action :create do
   file = win_friendly_path(@new_resource.source)
   isPfx = file.downcase.end_with?('pfx')
   password = ", \"#{@new_resource.pfx_password}\"" if isPfx
-  persistKeySet = ', ([System.Security.Cryptography.X509Certificates.X509KeyStorageFlags]::PersistKeySet)' if isPfx
+  persistKeySet = ', ([System.Security.Cryptography.X509Certificates.X509KeyStorageFlags]::PersistKeySet -bor [System.Security.Cryptography.X509Certificates.X509KeyStorageFlags]::MachineKeyset)' if isPfx
 
   codeScript = <<-EOH
     $store = New-Object System.Security.Cryptography.X509Certificates.X509Store "#{@new_resource.store_name}", ([System.Security.Cryptography.X509Certificates.StoreLocation]::#{@location})
--- a/providers/certificate.rb
+++ b/providers/certificate.rb
@@ -122,7 +122,7 @@ def load_current_resource
   @current_resource.private_key_acl(@new_resource.private_key_acl)
   @current_resource.store_name(@new_resource.store_name)
   @current_resource.user_store(@new_resource.user_store)
-  @location = @current_resource.user_store ? 'CurrrentUser' : 'LocalMachine'
+  @location = @current_resource.user_store ? 'CurrentUser' : 'LocalMachine'
 end
 
 private


chef-supermarket:
Hi. I am an automated pull request bot named Curry. There are commits in this pull request whose authors are not yet authorized to contribute to Chef Software, Inc. projects or are using a non-GitHub verified email address. To become authorized to contribute, you will need to sign the Contributor License Agreement (CLA) as an individual or on behalf of your company. [You can read more on Chef's blog.](https://www.chef.io/blog/2014/06/23/changes-to-the-contributor-license-agreement-process)

## Non-GitHub Verified Committers

There are 1 commit author(s) whose commits are authored by a non-GitHub verified email address. Chef will have to manually verify that they are authorized to contribute.

[Please sign the CLA here.](https://supermarket.chef.io/become-a-contributor)

tophercullen:
Should be verified now

ksubrama:
I dug into it and it seems https://support.microsoft.com/en-us/kb/950090 has a decent explanation of why that flag is needed.

Your patch requests that the private key go to the machine store regardless of where the cert goes.  Could you change it so you only append that flag to persistKeySet if new_resource.user_store is false?

tophercullen:
I just tested and confirmed the following: 
 1. Using the user store has the same problem when executing remotely (Access denied). 
 2. MachineKeySet doesn't seem to change the location the cert ends up in. Setting user_store to true puts the cert in the currentuser store.

I haven't done any testing with non-PFX certs, so I don't if they have the same problems, If so, the PersistKeySet will need to always be set, instead of just for PFX types.

ksubrama:
Out of curiosity, what version of windows are you running these tests from.  I happened to find https://support.microsoft.com/en-us/kb/2659071 .  Can you run winrm id on client and server and paste the ProductVersion and Stack?

tophercullen:
I'm using the latest WIndows server 2012 R2 AMI in EC2,  ProductVersion = OS: 6.3.9600 SP: 0.0 Stack: 3.0

ksubrama:
I revisited this today again - I'm in your camp now.  I can't seem to find a way to do this any other way.  Merging it now.

jdmundrawala:
Is this putting things in a place where anyone with access to the machine can see it?

tophercullen:
The key stores are still discrete (user vs machine). I don't know how/if this changes anything with how its stored to disk. My limited testing indicated the certs were in the normal paths for machine vs user key store.

ksubrama:
Anyone with admin.  It puts things in C:\ProgramData\Microsoft\Crypto\RSA\MachineKeys.

jdmundrawala:
https://msdn.microsoft.com/en-us/library/system.security.cryptography.x509certificates.x509keystorageflags%28v=vs.110%29.aspx

MachineKeySet	Private keys are stored in the local computer store rather than the current user store.

ksubrama:
The certs themselves are stored in the right place but the private keys are placed in the machine store.  I tried to puzzle through what the effective difference here (the certificate in a local store with the private key in the machine store) but it came down to one of it works with winrm remotely and the other doesn't and I'm not sure why...

PR #238, opened 2015-06-30
Don't always store private keys in machine store.

ksubrama:


ksubrama:
/cc @smurawski @jaym @btm 

This partially revert that key installation issue that made you guys nervous.

jaym:
Don't you need to modify the resource as well?

providers/certificate.rb
```diff
@@ -33,9 +33,13 @@ def whyrun_supported?
...
   # We can do everything in a powershell script resource
   file = win_friendly_path(@new_resource.source)
-  isPfx = file.downcase.end_with?('pfx')
-  password = ", \"#{@new_resource.pfx_password}\"" if isPfx
-  persistKeySet = ', ([System.Security.Cryptography.X509Certificates.X509KeyStorageFlags]::PersistKeySet -bor [System.Security.Cryptography.X509Certificates.X509KeyStorageFlags]::MachineKeyset)' if isPfx
+  password, persistKeySet = nil, nil
+  if file.downcase.end_with?('pfx')
+    password = ", \"#{@new_resource.pfx_password}\""
+    persistKeySet = '[System.Security.Cryptography.X509Certificates.X509KeyStorageFlags]::PersistKeySet'
+    persistKeySet = "(#{persistKeySet} -bor [System.Security.Cryptography.X509Certificates.X509KeyStorageFlags]::MachineKeyset)" unless @new_resource.user_store
```
jaym: This is some really weird code:
```
persistKeySet = ...
persistKeySet = ...
persistKeySet = ...
```

ksubrama:
Nope - https://github.com/opscode-cookbooks/windows/pull/229 was the affecting change.  The resource supports installing certs in the local machine or user stores.  The problems only arise when using the user stores remotely (and the original issue was that we missed a flag that caused the private key for a pfx requested to be held in the machine store to be stored in the user private key store).  This changes it so that we put private keys in the local machine private key store *only* for requests where the cert is imported into the local machine store.  That should work over winrm and remoting.

providers/certificate.rb
ksubrama: Yeaaah...  ok, let me clean it up with slightly better names.

ksubrama:
@jaym ZOMG refactor + some tests to prove that it actually works in the basic cases.  I hand verified the other cases because of test-kitchen winrm user profile loading issues.

smurawski:
@ksubrama - you could try the scheduled task runner with chef-zero-scheduled-task to get around the "running in winrm" issue.

ksubrama:
@smurawski True...  will do that for another PR.

jaym:
Awesome :+1:

PR #241, opened 2015-07-03
Adds supermarket tags

spuder:
Its too easy to misconfigure berks to pull from HEAD from master (when you think you are pulling a specific version)

https://github.com/opscode-cookbooks/windows/issues/240  
https://github.com/opscode-cookbooks/windows/issues/237  
https://github.com/opscode-cookbooks/windows/issues/90#issuecomment-49062111

The best solution would be to keep the master branch clean until the version is bumped. (Use a dev branch for accepting merge requests). 

This adds icons as a half measure to make it easier to identify what versions of a cookbook are available.


tas50:
I've added testing / cookbook version badges here

https://github.com/opscode-cookbooks/windows/commit/6c6053e7aad46063efb088d97adf1ebc2be7244b

PR #244, opened 2015-07-16
Add check for cert existence before binding

### LINKED ISSUE #230: windows_certificate_binding fails ungracefully if cert doesn't exist
Currently, when windows_certificate_binding resource fails due to a non-existent cert, it puts out a confusing windows error about sessions. The resource should check for the certs existence and, if it doesn't exist, fail with something more descriptive.

tophercullen:
If a cert doesn't exist, it shouldn't try and bind it.
Display a better error message than netsh would give
Fixes #230


ksubrama:
:+1: Thank you!

PR #246, opened 2015-07-23
Added windows_link and windows_service resource/providers

Chhed13:
Added extension to windows_service core resource/provider with ability to create/delete and fully config windows service
Added extension to core link resource/provider with ability to manipulate junction links (were only symbolic and hard)


Chhed13:
Hello, is any chance for merge? Or maybe it's better to move this request to chef-client core resources because I expand they current functionallity

mwrock:
I have not looked in detail, but I am curious why you decided to implement a `windows_service` resource here when one already exists in chef.

Chhed13:
Chef does'n have enough actions and attributes to cover all needs, with this you can do much more things. And I have to put it here because I use win32/service library which this cookbook install.
With *windows_link* situation is the same

Annih:
I really like the `config`, `create`, `delete` actions of this windows_service :)
The code is a bit complex to read, and sadly does not support `why_run`.

For the `win32-service` gem, [chef embedded it](https://github.com/chef/chef/blob/master/chef-universal-mingw32.gemspec#L16), you could move your code into chef core and fix https://github.com/chef/chef/issues/2518

:)

axelrtgs:
Would be nice to see this included in chef. i accidentally installed chef service on windows without realizing its deprecated and now although i want to keep chef i cannot remove the service using chef.

This functionality would be nice to be able to remove services for apps that are installed but that you do not want them to run as a service or to run an application as a service that doesn't install by default that way (such as a portable app)

tas50:
We've added the service resource logic to chef itself so at this point I'm going to close this PR out. If you'd like to open a PR up for just the link resource I'd be fine with that. That's probably a useful resource for a good number of users.

PR #253, opened 2015-07-28
Check if respond_to? source_url and issues_url

### LINKED ISSUE #252: 1.3.8 release breaks AWS opsworks
The 1.3.8 release includes the source_url directive in metadata.rb.  AWS opsworks only supports up to chef 11.10 on linux stacks, and the elastic search recipe depends on ark, which depends on windows and pulls in the latest version.  

Here is the error message from opsworks:

ERROR: Could not read /opt/aws/opsworks/current/merged-cookbooks/windows into a Chef object: undefined method `source_url' for #<Chef::Cookbook::Metadata:0x007febed0115e0>

The 1.3.7 release works fine.

Edit:  The source_url parameter is in the metadata.json file in 1.37.0 also, but there is no metadata.rb file in that version (at least when pulled into opsworks via Berkshelf) so the difference appears to be including that parameter in metadata.rb.

jaym:
Fixes #252

### DIFF
--- a/metadata.rb
+++ b/metadata.rb
@@ -6,6 +6,6 @@ description      'Provides a set of useful Windows-specific primitives.'
 long_description IO.read(File.join(File.dirname(__FILE__), 'README.md'))
 version          '1.38.1'
 supports         'windows'
-source_url       "https://github.com/opscode-cookbooks/windows"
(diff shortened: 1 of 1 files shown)


JamesMGreene:
:+1: X :100:

smurawski:
:+1:

PR #257, opened 2015-08-01
Speed up windows_feature dism provider

Annih:
Actions :enable & :disable check the feature's state using dism.exe.
It's a bit slow, and can be easily replaced by a simple wmi query.

cc. @aboten


mwrock:
:+1: 

smurawski:
LGTM :+1:

PR #275, opened 2015-09-04
Removes documentation of CWD since its broken

spuder:
https://github.com/opscode-cookbooks/windows/issues/176

Removes cwd so that users aren't mistaken into believing that it works.


spuder:
Anything stopping this from being accepted?

mwrock:
I think this should be fixed ratrher than moving the references in the docs. If we were going to simply remove the feature, we would also want to remove the attribute from the resource and provider.

mwrock:
Just submitted a fix via #311

Issue #276, opened 2015-09-09
LoadError: windows/error


mcquin:
Unable to compile the windows cookbook in a recipe which depends on it.

System: win2012r2-x86_64
Chef client version: 12.5.0.current.0 (64-bit)

Error:
```
LoadError
-------------
cannot load such file -- windows/error
```
Cause: https://github.com/chef-cookbooks/windows/blob/master/libraries/windows_privileged.rb#L24

Repro: 
```
# Clone this repository and cd to it
git clone https://github.com/adamedx/winbox
cd winbox

# Use Berkshelf to download all cookbook dependencies:
berks vendor cookbooks

# Run `chef-client` in local mode to configure the workstation!
chef-client -z --color -o 'winbox'
```

mcquin:
It looks like `windows/*` comes from the `windows-pr` gem, which is being removed from Chef.

rcsavage:
Confirmed, I ran into this problem as well.   I simply went into C:\opscode\Chef\embedded\bin, and then a gem install windows-pr and that fixed the problem.   Thank you mcquin for finding this.

jfbibeau:
Sorry to comment on a closed issue, but does this mean this cookbook isn't compatible with Chef 12.5.1 until the commit referenced above goes in a cookbook release?

mcquin:
@jfbibeau It looks like version 1.38.2 of this cookbook is released on Supermarket. The fix is included in 1.38.2, so if you update your windows cookbook dependency it should work with Chef 12.5+.

jfbibeau:
@mcquin Of course, sorry I missed that. 1.38.2 fixed the issue. Cheers.

JackChance:
still an issue with windows 1.44.3

mwrock:
what version of the chef client @JackChance ?

JackChance:
12.6.0



Sent from my Verizon 4G LTE smartphone


-------- Original message --------

JackChance:
Sorry, let me add more details. I'm bootstrapping from OEL 6.5 to win2012r2.

OEL 6.5 bootstrapper has ChefDK 0.13.21 which includes chef-client 12.9.41
win2012r2 bootstrappee has chef-client 12.6.0.

Bootstrap succeeds in registering to chef server and installing client, fails when it attempts to include windows/errors. Attempted to `chef gem install windows-pr` also errors out, with well over a hundred errors..

```
make "DESTDIR="
compiling win32/api.c
[...]

PR #277, opened 2015-09-11
Lazy load windows-pr dependency

mcquin:
Resolves https://github.com/chef-cookbooks/windows/issues/276

Chef (client) 12.5 no longer ships with the windows-pr gem. This cookbook includes the windows-pr gem but the windows/* libraries were being loaded from the windows-pr gem shipping with Chef. This lazy-loads the windows/* libraries so the cookbook can compile.

### DIFF
--- a/libraries/windows_privileged.rb
+++ b/libraries/windows_privileged.rb
@@ -20,24 +20,14 @@
 # limitations under the License.
 #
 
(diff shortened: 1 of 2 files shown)


mcquin:
indenting is stupid.

PR #278, opened 2015-09-11
Lazy load windows-pr

mcquin:
Resolves https://github.com/chef-cookbooks/windows/issues/276

Lazy loads windows-pr libraries. Chef 12.5 no longer includes the windows-pr gem, which is where these files were actually being loaded from (not from the `chef_gem 'windows-pr`').

### DIFF
--- a/libraries/windows_privileged.rb
+++ b/libraries/windows_privileged.rb
@@ -20,24 +20,13 @@
 # limitations under the License.
 #
 
-if RUBY_PLATFORM =~ /mswin|mingw32|windows/
-  require 'windows/error'
-  require 'windows/registry'
-  require 'windows/process'
-  require 'windows/security'
-end
-
 #helpers for Windows API calls that require privilege adjustments
 class Chef
   class WindowsPrivileged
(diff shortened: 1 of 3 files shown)


jaym:
:+1:

libraries/windows_privileged.rb
```diff
@@ -20,24 +20,13 @@
...
   class WindowsPrivileged
-    if RUBY_PLATFORM =~ /mswin|mingw32|windows/
-      include Windows::Error
-      include Windows::Registry
-      include Windows::Process
-      include Windows::Security
-    end
     #File -> Load Hive... in regedit.exe
     def reg_load_key(name, file)
+      load_deps
```
chefsalim: Is there an additional cost to re-require and include on every call?
If so, do something along the lines of @loaded ||= load_deps ?

mcquin: There may be, but it won't be much. I know that ruby doesn't reload a file the second time it is required.

chefsalim: OK then LGTM :+1:

PR #284, opened 2015-09-24
Resolve FC001 warnings in providers

tas50:


tas50:
So this gets us to a clean Foodcritic run.  The current cookbook in the Supermarket fails foodcritic for FC001 which puts a big red X on the supermarket download page.  Additional we've identified passing all Foodcritic rules as a metric of cookbook quality in our current cookbook quality metrics RFC.

https://github.com/chef-cookbooks/cookbook-quality-metrics/blob/master/quality-metrics/qm-013-foodcritic.md

jaym:
ok :+1: 
Do you have a reference as to why this rule is around?

tas50:
There's a large number of conversations on FC001.  The rule was pulled at one point and then added back at the request of many.  The current recommended best practices for the community are to follow FC001 so users should be using strings.

https://www.google.com/search?q=fc001&oq=fc001&aqs=chrome..69i57j69i61j0l4.1199j0j7&sourceid=chrome&es_sm=91&ie=UTF-8#q=fc001+foodcritic

PR #286, opened 2015-09-28
frequency_modifier validation & additional parameters

jboeshart:
I've updated the windows_task provider to validate the `frequency_modifier` attribute to ensure it's a valid value and updated the same resource to allow Strings. This allows you to specify the additional parameters for the `schtasks.exe` switch `/MO` that are string values (LAST, FIRST, SECOND, THIRD, FOURTH, & LAST). 

Feedback welcome on the code if there is a more elegant way to handle this or if I missed anything.


mwrock:
Looks good but looks like #305 supersedes this one.

PR #294, opened 2015-11-04
Auto correct rubocop warnings and disable others

tas50:
Get Travis back to green.  Lots of automatic changes here.


philoserf:
Why not use the TODO provided by the rubocop tool ala `rubocop --auto-gen-config`?

Issue #298, opened 2015-11-16
windows_certificate_binding doesn't follow store name


pvandervelde:
It seems that the windows_certificate_binding ignores the provided store name and always uses the MY certificate store.
My cookbook entry looks like:

    portal_host_name = 'aktfsjs01'
    certificate_file_name = portal_host_name.gsub('.', '_').gsub('*', '')
    ssl_certificate_password = node['password']['ssl_certificate']
    windows_certificate "c:/temp/configuration/cookbooks/mobile_portal/files/windows/#{certificate_file_name}.pfx" do
      pfx_password ssl_certificate_password
      store_name 'CA'
    end

    windows_certificate_binding portal_host_name do
      store_name 'CA'
      port 443
    end

This correctly installs the certificate in the CA store (though I would love to use the WebHosting store), but it doesn't bind the certificate to the correct port.

In the log I get

    [2015-11-16T18:00:06+13:00] INFO: Processing windows_certificate_binding[aktfsjs01] action create (mobile_portal::portal line 168)
    [2015-11-16T18:00:07+13:00] DEBUG: netsh reports: 

    SSL Certificate bindings: 

    ------------------------- 

    The system cannot find the file specified.

    [2015-11-16T18:00:07+13:00] DEBUG: Running PS script & { gci cert:\localmachine\MY | where subject -like '*aktfsjs01*' | select -first 1 -expandproperty Thumbprint }

And it's of course correct that it can't find the certificate because it is looking in the MY store but I installed it in the CA store (and I'm telling it that it's in the CA store too!).

Unfortunately I can't install in the MY store because the user account that is used to do the configuration is not the same as the user that will be used to run the web site that needs the certificate.

PR #299, opened 2015-11-17
windows_certificate_binding is ignoring store_name attribute

mwrock:
This fixes issues as reported in #298 where the `store_name` of `windows_certificate_binding` is ignored and always uses `MY` regardless of attribute value.


jaym:
:+1:

btm:
:+1:

PR #302, opened 2015-11-17
fixes certificate idempotentcy on chef 11 clients

### LINKED ISSUE #296: windows_certificate create action does not appear to be idempotent
Converging on a Windows 2012 RTM virtual box the windows_certificate resource always runs irrespective of whether the certificate is installed in the certificate store.

Example usage of the windows_certificate resource
```
  windows_certificate "C:/certs/my_cert_file" do
    pfx_password 'my_password'
  end  
```
Having looked at the code I think the powershell resource used in the create method needs changing to specify the guard interpreter 
```
  powershell_script new_resource.name do
    guard_interpreter :powershell_script
    code code_script
    not_if guard_script
  end
```
Alternatively if a guard_interpreter is not specified the resource can be made idempotent by returning an exit code of 0 or 1 from the script by changing it to
```
def cert_exists_script(hash)
  <<-EOH
$hash = #{hash}
If ((Test-Path "Cert:\\#{@location}\\#{new_resource.store_name}\\$hash") -eq $true) {exit 0} else {exit 1}
EOH
end
```

Both solutions above tested with Chef 11.18.6 and 12.4.1.

mwrock:
fixes #296

### DIFF
--- a/providers/certificate.rb
+++ b/providers/certificate.rb
@@ -39,7 +39,9 @@ action :create do
   guard_script = cert_script(false) <<
                  cert_exists_script(hash)
 
-  powershell_script new_resource.name do
+  powershell_script @new_resource.name do
+    guard_interpreter :powershell_script
+    convert_boolean_return true
     code code_script
     not_if guard_script
   end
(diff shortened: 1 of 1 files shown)


chefsalim:
Fix looks good to me.. minor comment - should the new_resource var be changed to @new_resource? looks like both new_resource and @new_resource are being used.

smurawski:
:+1:  other than the same minor nitpick as @chefsalim

mwrock:
ok i changed the `new_resource` variable scoping.

PR #295, opened 2015-11-06
Escape http_acl user in shellout

Annih:
Current `http_acl` implementation fail when username contains space(s).
Surrounding the netsh param with double quotes fix the issue.

cc. @aboten

PR #303, opened 2015-11-18
Fix http_acl regex to capture username with space

Annih:
With #295 http_acl LWRP now supports username with spaces, but due to the current regex to capture username, the execution is not idempotent.
Changing the regex to capture everything after space characters fix it.

cc: @aboten

### DIFF
--- a/providers/http_acl.rb
+++ b/providers/http_acl.rb
@@ -73,7 +73,7 @@ def getCurrentAcl
   cmd = shell_out!("#{@command} http show urlacl url=#{@current_resource.url}")
   Chef::Log.debug "netsh reports: #{cmd.stdout}"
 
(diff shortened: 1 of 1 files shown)


mwrock:
:+1: 

smurawski:
:+1: Would be nice to get a test for this as well, but that shouldn't block merging.

Annih:
@smurawski  I'm working on a tiny new feature for this LWRP and I'll provide some test cases

mwrock:
I also added a test for basic functionality https://github.com/chef-cookbooks/windows/commit/bf51eb3a1f81bc768ba28ddeabb9120cd7c1dbcb

mwrock:
We are looking to relase late AM today Pacific time.

PR #305, opened 2015-11-18
Added Months option and added frequency_modifier def to handle both i…

### LINKED ISSUE #304: windows_task support for Months and frequency modifier strings for first, second, etc...
windows_task needs support for months (schtasks /create /M) and frequency modifier to accept strings for (first, second, third, etc...). This allows companies like ours to create tasks for scheduling stuff like patches on 3rd tuesday of all months except december due to the holidays.

Stromweld:
Added Months option and added frequency_modifier def to handle both integers and strings. Fixes #304

### DIFF
--- a/providers/task.rb
+++ b/providers/task.rb
@@ -237,42 +237,43 @@ end
 def validate_create_months
   return unless @new_resource.months
   unless [:monthly].include?(@new_resource.frequency)
-    fail 'months attribute is only valid for tasks that run weekly or monthly'
+    fail 'months attribute is only valid for tasks that run monthly'
   end
-  months = @new_resource.months.split(',')
-  months.each do |month|
-    unless ['JAN', 'FEB', 'MAR', 'APR', 'MAY', 'JUN', 'JUL', 'AUG', 'SEP', 'OCT', 'NOV', 'DEC', '*'].include?(month.upcase)
-      fail 'months attribute invalid. Only valid values are: JAN, FEB, MAR, APR, MAY, JUN, JUL, AUG, SEP, OCT, NOV, DEC and *. Multiple values must be separated by a comma.'
+  if @new_resource.months.is_a? String
+    months = @new_resource.months.split(',')
+    months.each do |month|
+      unless ['JAN', 'FEB', 'MAR', 'APR', 'MAY', 'JUN', 'JUL', 'AUG', 'SEP', 'OCT', 'NOV', 'DEC', '*'].include?(month.upcase)
+        fail 'months attribute invalid. Only valid values are: JAN, FEB, MAR, APR, MAY, JUN, JUL, AUG, SEP, OCT, NOV, DEC and *. Multiple values must be separated by a comma.'
+      end
     end
   end
 end
 
 def validate_create_frequency_modifier
+  # Currently is handled in create action 'frequency_modifier_allowed' line. Does not allow for frequency_modifier for once,onstart,onlogon,onidle
+  # Note that 'OnEvent' is not a supported frequency.
   unless @new_resource.frequency.nil? || @new_resource.frequency_modifier.nil?
     case @new_resource.frequency
-      when :minute
-        unless @new_resource.frequency_modifier.to_i > 0 && @new_resource.frequency_modifier.to_i <= 1439
-          fail "frequency_modifier value #{@new_resource.frequency_modifier} is invalid.  Valid values for :minute frequency are 1 - 1439."
-        end
-      when :hourly
-        unless @new_resource.frequency_modifier.to_i > 0 && @new_resource.frequency_modifier.to_i <= 23
-          fail "frequency_modifier value #{@new_resource.frequency_modifier} is invalid.  Valid values for :hourly frequency are 1 - 23."
-        end
-      when :daily
-        unless @new_resource.frequency_modifier.to_i > 0 && @new_resource.frequency_modifier.to_i <= 365
-          fail "frequency_modifier value #{@new_resource.frequency_modifier} is invalid.  Valid values for :daily frequency are 1 - 365."
-        end
-      when :weekly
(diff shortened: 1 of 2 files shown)


Stromweld:
the rubocop failures for case where it wants the when statements indented as deep as the case statement, is that how it should be. My RubyMine IDE automagically indents them one level deeper and readability I think that it's nice as you can tell they are a condition of the case statement.

providers/task.rb
```diff
@@ -231,6 +234,49 @@ def validate_create_day
   end
 end
 
+def validate_create_months
+  return unless @new_resource.months
+  unless [:monthly].include?(@new_resource.frequency)
+    fail 'months attribute is only valid for tasks that run weekly or monthly'
```
mwrock: change message to just monthly

providers/task.rb
```diff
@@ -231,6 +234,49 @@ def validate_create_day
...
+  end
+  months = @new_resource.months.split(',')
+  months.each do |month|
+    unless ['JAN', 'FEB', 'MAR', 'APR', 'MAY', 'JUN', 'JUL', 'AUG', 'SEP', 'OCT', 'NOV', 'DEC', '*'].include?(month.upcase)
+      fail 'months attribute invalid. Only valid values are: JAN, FEB, MAR, APR, MAY, JUN, JUL, AUG, SEP, OCT, NOV, DEC and *. Multiple values must be separated by a comma.'
```
mwrock: i think you will want to `chomp` in case there are spaces between commas. But even better specify `Array` as the `kind_of` for this attribute.

providers/task.rb
```diff
@@ -231,6 +234,49 @@ def validate_create_day
...
+      when :weekly
+        unless @new_resource.frequency_modifier.to_i > 0 && @new_resource.frequency_modifier.to_i <= 52
+          fail "frequency_modifier value #{@new_resource.frequency_modifier} is invalid.  Valid values for :weekly frequency are 1 - 52."
+        end
+      when :monthly
+        unless ('1'..'12').to_a.push('FIRST', 'SECOND', 'THIRD', 'FOURTH', 'LAST', 'LASTDAY').include?(@new_resource.frequency_modifier.to_s.upcase)
+          fail "frequency_modifier value #{@new_resource.frequency_modifier} is invalid.  Valid values for :monthly frequency are 1 - 12, 'FIRST', 'SECOND', 'THIRD', 'FOURTH', 'LAST', 'LASTDAY'."
+        end
+      else
+        # Currently is handled in create action 'frequency_modifier_allowed' line. Does not allow for frequency_modifier for once,onstart,onlogon,onidle
```
mwrock: I'd put this comment in the top of the method and ommit the `else`

resources/task.rb
```diff
@@ -43,6 +43,7 @@
 attribute :start_day, kind_of: String, default: nil
 attribute :start_time, kind_of: String, default: nil
 attribute :day, kind_of: [String, Integer], default: nil
+attribute :months, kind_of: String, default: nil
```
mwrock: consider `Array` instead of `String`

mwrock:
For the sake of consistency, we'd want this to comply with the rubocop rules currently used. We don't want to merge a PR unless it is green.

mwrock:
Oh and I should also say thank you very much for this! I think it is a good addition.

providers/task.rb
Stromweld: I can certainly do that, but right now it's consistent with the rest of the file and follows same format as "def validate_create_day". Maybe another revision could replace them both with arrays.

resources/task.rb
Stromweld: I was mainly trying to stay consistent with the way the day frequency was written.

Stromweld:
Made the rubocop fixes and a couple of your recommendations from comments above. As far as array's go, that would probably be best but not sure if that should be in this scope or a refactor afterwards. I was trying to stay consistent with the rest of the file and formatted it like how validate_create_day is. 

I am also fairly new to ruby, chef, and programming in general and have been learning as I go I'm not totally sure what would all need to be done to refactor for kind_of array.

resources/task.rb
mwrock: Yeah I took a look at the windows `schtasks` docs and I think day is meant to take just one day: 
```
    /D    days         Specifies the day of the week to run the task. Valid
                       values: MON, TUE, WED, THU, FRI, SAT, SUN and for
                       MONTHLY schedules 1 - 31 (days of the month).
                       Wildcard "*" specifies all days.

    /M    months       Specifies month(s) of the year. Defaults to the first
                       day of the month. Valid values: JAN, FEB, MAR, APR,
                       MAY, JUN, JUL, AUG, SEP, OCT, NOV, DEC. Wildcard "*"
                       specifies all months.
```
The `/M` flag mentions "month(s)" while `/D` just says "day".

mwrock:
Yeah the Array is not a big deal. I think it might make things easier but wouldn't hold off a merge. However, after you split the string on ',' you should `strip` the individual months. Otherwise if someone specifies:
```
'JAN, FEB'
```
FEB will be ' FEB'

resources/task.rb
Stromweld: actually looking at it from the gui side task scheduler, with the weekly and monthly settings you can specify multiple days. Weekly gives you mon,tue,wed,etc... options and monthly days gives you the 1-31 etc... options

mwrock: oh ok.

mwrock: still needs a `strip` though: `.include?(month.upcase.strip)`

Stromweld:
is this last commit what your talking about?

mwrock:
yeah

Stromweld:
cool

mwrock:
:+1:  cc @chef-cookbooks/client-windows @chef/client-windows

adamedx:
:+1:

PR #314, opened 2015-12-10
fixes version comparison bug

andyregan:
Commit 7471d87e6ca4e79eeffa67cb540669d912071852 introduced a change to use Chef DSL
if the chef-client version is greater than 11.12. However, since it compares strings
the results are unexpected (for example "11.6.2" is greater than "11.12").

This change compares Gem::Version objects instead of strings.

(Thanks Jon Lenzer :D )


mwrock:
:+1: 

cc @chef-cookbooks/client-windows @chef/windows-client

jaym:
:+1:

mwrock:
Thanks @andyregan !

andyregan:
Thanks!

PR #315, opened 2015-12-14
deprecate windows_package and forward to new provider if running 12.6 or higher

mwrock:


chefsalim:
LGTM :+1:

Tech356:
If the windows cookbook is included for any reason, won't this throw the warning every time? 

Since the Chef Core version is named `windows_package` just like the cookbook `windows_package`, the cookbook version will override the core version and cause the warning to be thrown? The only way to get rid of the warning would be to remove the windows cookbook from the run list so the core version of `windows_package` doesn't get overridden, correct?

mwrock:
@Tech356 you could get around it by using `package` instead of `windows_package`. On a windows system that will use the core `windows_package` without going through the cookbook.

smurawski:
:+1:

PR #318, opened 2015-12-16
Use dsl instead of manual resource instanciation

Annih:
Some providers and libraries are manually instanciating resources.
This is not really clean, and bypass many steps of the resource
initialization - setting cookbook name for instance.

I use the DSL and manually call `run_action` when it's required.

cc: @aboten


mwrock:
:+1: 

smurawski:
:+1: :shipit:

PR #333, opened 2016-01-27
Create feature_cached.rb

mattchurchy:
This is a cached version of feature_dism. Dism is ran once and a copy is stored in memory and reused for the life time of chef-client. As such it is an order of magnitude quick when checking multiple features
The cache is cleared if pages are installed or removed via this same provisioner.

For now to use set node.default['windows']['feature_provider'] = 'cached'.
Perhaps once it's proven it can be added to feature.rb as a normal option.

I've tested this with the provided test kitchen resources as well as with our own cookbooks.

### DIFF
--- a/providers/feature_cached.rb
+++ b/providers/feature_cached.rb
@@ -1,0 +1,80 @@
+#
+# Author:: Matthew Churcher (<Matthew.Churcher@newvoicemedia.com>)
+# Heavily based on feature_dism.rb by Seth Chisamore (<schisamo@chef.io>)
+# Cookbook Name:: windows
+# Provider:: feature_cached
+#
+# Copyright:: 2016, Chef Software, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+include Chef::Provider::WindowsFeature::Base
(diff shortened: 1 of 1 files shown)


chef-supermarket:
Hi. I am an automated pull request bot named Curry. There are commits in this pull request whose authors are not yet authorized to contribute to Chef Software, Inc. projects or are using a non-GitHub verified email address. To become authorized to contribute, you will need to sign the Contributor License Agreement (CLA) as an individual or on behalf of your company. [You can read more on Chef's blog.](https://www.getchef.com/blog/2014/06/23/changes-to-the-contributor-license-agreement-process)

## GitHub Users Who Are Not Authorized To Contribute

The following GitHub users do not appear to have signed a CLA:

* @mattchurchy

[Please sign the CLA here.](https://supermarket.chef.io/become-a-contributor)

mwrock:
This is a great idea and I'd be inclined to say go ahead and just add to the dism feature. The unfortunate thing is that if you mix feature providers in a single run, the cache could be invalid. Its certainly possible to mix providers but unlikely unless you have multiple community cookbooks wrapping `windows_feature` and using different providers.

matthew-churcher:
Same as https://github.com/chef-cookbooks/windows/pull/339. Just trying to the CLA signing recognised. Also just saw @mwrock comments so will give that some thought.

smurawski:
@mattchurchy Regarding the CLA - is the email on your commit (your gmail address) associated with your github user account?  And that github user account associated with your Chef supermarket account? 

As to the technical aspect, I like Matt's thinking and making sure that the cache is only hit for like providers.

mattchurchy:
Whoa CCLA should be sorted now! I'll have to switch to my other git-hub account though.

Sorry but I'm not following how you want to proceed....
You want me to modify Feature_Dism and not worry about other providers?
 
I could introduce a new attribute as a fail-safe to allow caching to be disabled?

Regardless of whether multiple providers is handled, there will remain the possibility of software installing windows features in the background. This feels like a niche scenario to me though, and a user could just converge again. So having an attribute feels like a reasonable compromise to me for users to need to handle that scenario.

PR #341, opened 2016-02-13
fix rubcop warnings

mwrock:


libraries/windows_package.rb
```diff
@@ -7,10 +7,12 @@
 require 'chef/mixin/language'
 class Chef
   class Provider
-    class WindowsCookbookPackage < Chef::Provider::LWRPBase
+    class WindowsCookbookPackage < Chef::Provider::LWRPBase # ~FC058
```
randomcamel: Why the comment?

randomcamel:
:+1:

libraries/windows_package.rb
mwrock: suppresses a foodcritic warning that just started with FC 6.0.0. Since this is a deprecated resource, its really not worth refactoring.

smurawski:
:+1:

PR #348, opened 2016-03-04
fixes #271 making notify work for windows_task

### LINKED ISSUE #271: windows_task will not 'notify' other resources
If you have a scheduled task, and you attempt to delete it with ':delayed' it will never be deleted. 

Example

```
  windows_task 'foobar' do
    user "example.com\\bob"
    command  "C:\\Windows\\System32\\WindowsPowerShell\\v1.0\\powershell.exe -noprofile -noexit -executionpolicy Bypass -File c:\foo.ps1 '"
    action [:create,:run]
    notifies :end, 'windows_task[foobar]', :delayed  #TODO This strangely never gets deleted
    notifies :delete, 'windows_task[foobar]', :delayed #TODO This strangely never gets deleted
    run_level :highest
    force true
  end
```

Windows 2012 R2
Chef 12.4.1
Windows cookbook 1.38.1

mwrock:


smurawski:
rubocop is not so happy :/

smurawski:
Once rubocop is happy :+1:

mwrock:
cc @chef-cookbooks/windows-cookbook-collaborators @chef-cookbooks/client-windows

adamedx:
what was root cause of bug -- not using a consistent name in different contexts?

mwrock:
the `.prepend` method on `task_name` changes the state of the string (should be a `prepend!` if you ask me but sadly they did not ask) so since that is assigned to the resource_name the resource_name changes and then the notify goes nowhere.

tyler-ball:
:+1:

PR #349, opened 2016-03-08
Update feature_dism with cache

matthew-churcher:
Add caching to feature_dism see #333 for more detail

### DIFF
--- a/providers/feature_dism.rb
+++ b/providers/feature_dism.rb
@@ -27,10 +27,12 @@ def install_feature(_name)
   addsource = @new_resource.source ? "/LimitAccess /Source:\"#{@new_resource.source}\"" : ''
   addall = @new_resource.all ? '/All' : ''
   shell_out!("#{dism} /online /enable-feature /featurename:#{@new_resource.feature_name} /norestart #{addsource} #{addall}", returns: [0, 42, 127, 3010])
+  clear_cache
 end
 
 def remove_feature(_name)
   shell_out!("#{dism} /online /disable-feature /featurename:#{@new_resource.feature_name} /norestart", returns: [0, 42, 127, 3010])
+  clear_cache
 end
 
 def delete_feature(_name)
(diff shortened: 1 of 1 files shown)


providers/feature_dism.rb
```diff
@@ -69,3 +67,14 @@ def dism
     locate_sysnative_cmd('dism.exe')
   end
 end
+
+@@feature_list = nil
+def get_feature_list
```
mwrock: Might be nice to have this create a hash of features to avoid the parsing in the consuming methods

mwrock:
Good to see this in `feature_dism`!

This came up before but I can see cases where consuming multiple windows based cookbooks, you will have recipes possibly checking features via different providers. So one could run into this sequence:
1. Hydrate the dism cache
2. Install a feature via powershell
3. Try to install the same feature here later in the run list.

I **think** this is an edge case and would likely just result in an unnecessary call to install the feature but not result in an error. So I think this is ok but if anyone sees a graver danger we should address it.

matthew-churcher:
@mwrock I made the suggested change, a hash is now used to store the features to avoid the parsing dism output again.

providers/feature_dism.rb
```diff
@@ -69,3 +68,22 @@ def dism
     locate_sysnative_cmd('dism.exe')
   end
 end
+@@feature_list = nil
+def get_feature_list
+  if @@feature_list.nil?
+    list = {}
+    out = shell_out("#{dism} /online /Get-Features", returns: [0, 42, 127]).stdout
```
mwrock: should this be `feature_matches` instead of `out`?

providers/feature_dism.rb
```diff
@@ -69,3 +68,22 @@ def dism
...
+    feature_matches.each do |match|
```
mwrock: could use:
`@@feature_list = feature_matches.each_with_object({}) do |match, list|`

providers/feature_dism.rb
matthew-churcher: Nice, spot. There is an entire line missing. Not sure how that made it through testing.

providers/feature_dism.rb
matthew-churcher: done

providers/feature_dism.rb
```diff
@@ -69,3 +68,22 @@ def dism
...
+  if @@feature_list.nil?
+    out = shell_out("#{dism} /online /Get-Features", returns: [0, 42, 127]).stdout
+    feature_matches = out.scan(/^Feature Name : (.*).?$\n^State : (.*)?$/)
+    list = feature_matches.each_with_object({}) do |match, list|
+      name = match[0]
+      state = match[1]
+      state = 'Removed' if state.include? 'Payload Removed'
+      list[name] = state.to_sym
+    end
+    @@feature_list = list
```
mwrock: just a nit: I'd lose this line and just assign `@@feature_list` to the block

mwrock:
:+1:

matthew-churcher:
@mwrock Thanks for the feed back. It's been a while since ruby was my goto language! Technically I'm a performance test engineer :) probably shouldn't be writing code at all, lol.

What are the prospects of getting this merged? the team at NVM is keen to get a speed up in this area but we would prefer not to maintain a fork. This patch speed's up each of our converges by several minutes. (yay)

matthew-churcher:
Oh deer, says I closed it. How to I reopen it?

mwrock:
cc @chef-cookbooks/windows-cookbook-collaborators @chef-cookbooks/client-windows

tas50:
The latest release of this cookbook includes caching via an Ohai plugin for dism. This should achieve the same thing and gives us a nice cache reloading method within ohai.

PR #350, opened 2016-03-08
Adding a vagrant user where needed in case we want to run this on AWS

tyler-ball:
With chef-acceptance we want to run this recipe on an AWS node, but the vagrant user doesn't exist there.  What do you think of this change @smurawski @mwrock ?  Or we could do a custom user like the [http_acl](https://github.com/chef-cookbooks/windows/blob/master/test/cookbooks/minimal/recipes/http_acl.rb) recipe uses

### DIFF
--- a/test/cookbooks/minimal/recipes/tasks.rb
+++ b/test/cookbooks/minimal/recipes/tasks.rb
@@ -59,6 +59,10 @@ windows_task 'delete task delete_me' do
   action :delete
 end
 
+user 'vagrant' do
+  password 'vagrant'
+end
+
 windows_task 'task_for_system' do
   command 'dir'
   run_level :highest


mwrock:
This recipe is really intended to be a spec for the `windows_task` resource. I think this is probably the wrong place for that. Maybe we could add a recipe to the beginning of a  chef-acceptance run that includes this.

smurawski:
That spec needs to change since it assumes vagrant.. maybe make the user for the task an attribute of the minimal cookbook?  It can default to vagrant, but then be overridable easily in a .kitchen.yml.  I've actually been meaning to make that change myself.

tyler-ball:
Closing this because we decided to add the `user 'vagrant'` into an acceptance specific recipe

PR #351, opened 2016-03-10
Add kitchen verify to Appveyor run

mwrock:
This runs all kitchen tests in appveyor. Should be fairly fast since it just uses the instance we already have. The motivation here is better feedback to contributors and less time in triage/testing.

### DIFF
--- a/.kitchen.appveyor.yml
+++ b/.kitchen.appveyor.yml
@@ -1,0 +1,48 @@
+---
+driver:
+  name: proxy
+  host: localhost
+  reset_command: "exit 0"
+  port: 5985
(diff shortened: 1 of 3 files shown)


mwrock:
consider this "In Progress" until green.

mwrock:
from what I can tell @smurawski the tests as they stand now should run on from all suites on a single node and not corrupt each other. Thats not to say that could not change but I think something to watch for. The nice thing is that since the appveyor VMs are ephemeral, at least each run should be fresh.

tas50:
:+1: DO IT!

PR #352, opened 2016-03-24
Resolve issue 330 (app_id not respected for certificate_binding)

### LINKED ISSUE #330: windows_certificate_binding always uses the default value for app_id
### Environment:

Windows 2012 R2, windows cookbook version 1.38.4
### Scenario:

When trying to bind a certificate to an app_id other than the default, it still binds to the default of `{4dc3e181-e14b-4a21-b022-59fc669b0914}`.
### Steps to Reproduce:

``` ruby
windows_certificate_binding 'bind the ssl certificate to port 443' do
    cert_name 'some hash'
    name_kind :hash
    app_id '{00000000-0000-0000-0000-000000000000}'
    port 443
    action :create
end
```
### Expected Result:

For the ssl cert to bind to the app_id passed into the resource.
### Actual Result:

It still binds to the default of `{4dc3e181-e14b-4a21-b022-59fc669b0914}`.
### Suspected Reason

app_id is not actually listened to, it does not get put into @current_resource https://github.com/chef-cookbooks/windows/blob/master/providers/certificate_binding.rb#L70 so line https://github.com/chef-cookbooks/windows/blob/master/providers/certificate_binding.rb#L98 always has default value of '{4dc3e181-e14b-4a21-b022-59fc669b0914}'

rlaveycal:
Fixes #330 

Also fixed weird issue with BOM being returned when getting hash from subject
Tests also need to use upper case hashes for paths (win 2008R2)

### DIFF
--- a/libraries/matchers.rb
+++ b/libraries/matchers.rb
@@ -7,6 +7,7 @@ if defined?(ChefSpec)
   end
 
   define_method.call :windows_certificate
+  define_method.call :windows_certificate_binding
   define_method.call :windows_package
   define_method.call :windows_feature
   define_method.call :windows_task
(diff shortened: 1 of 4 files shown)


mwrock:
At a glance this looks good. Just fix up the minor rubocop violations and lets see this go green.

tas50:
I manually merged this PR

PR #355, opened 2016-04-02
bump and pin rubocop and fix broken cop

mwrock:
PRs are starting to break on rubocop 0.39.2. This fixes violations and pins to that.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -7,7 +7,7 @@ end
 
 group :lint do
   gem 'foodcritic', '~> 6.0'
-  gem 'rubocop', '~> 0.34'
+  gem 'rubocop', '~> 0.37.2'
 end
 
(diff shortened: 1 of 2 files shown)


tas50:
You probably want to pin to the version included in ChefDK with Chefstyle now, which is 0.38 I believe.

mwrock:
Sorry the above was a typo. It's 0.37.2 which is the exact same as what is in chef-dk master. https://github.com/chef/chef-dk/blob/master/omnibus/config/projects/chefdk.rb#L47

tas50:
@mwrock Oddly enough ChefDK runs in Travis say it's 0.38, but on my laptop is says 0.37.2. :+1: this though

smurawski:
:+1:

PR #356, opened 2016-04-04
Misc fixes and updates

tas50:


chefignore
```diff
@@ -93,9 +100,3 @@ Strainerfile
 ###########
 .vagrant
 Vagrantfile
-
-# Travis #
-##########
-.travis.yml
-
-.kitchen
```
mwrock: we want to keep .kitchen here otherwise it gets included in the berks install and if you use Hyper-V (like me) that includes the test image file.

mwrock:
I'm :+1: as long as we put .kitchen back in chefignore and get a green appveyor

mwrock:
If you rebase on master now, the style should be good.

tas50:
@mwrock .kitchen* is still in there

PR #357, opened 2016-04-06
Some fixes for schtasks.exe parameters (on_idle parameters, day can be a number, validations)

elpetak:
Fixes on_idle which should be onidle in schtasks.exe command line.
Adds frequency_modifier interpretation for on_idle (/I option is required for onidle)
Fixes /MO being included in some cases when it shouldn't be.
Fixes day validation when day is a number (used in some monthly variants).
Fixes frequency_modifier validation for on_idle.


providers/task.rb
```diff
@@ -36,12 +36,22 @@
     validate_create_day
     validate_create_months
 
-    schedule = @new_resource.frequency == :on_logon ? 'ONLOGON' : @new_resource.frequency
+    schedule = case @new_resource.frequency
```
mwrock: I'd extract out to a method

providers/task.rb
```diff
@@ -36,12 +36,22 @@
...
+                 'ONIDLE'
+               else
+                 @new_resource.frequency
+               end
     frequency_modifier_allowed = [:minute, :hourly, :daily, :weekly, :monthly]
     options = {}
     options['F'] = '' if @new_resource.force || task_need_update?
     options['SC'] = schedule
-    options['MO'] = @new_resource.frequency_modifier if frequency_modifier_allowed.include?(@new_resource.frequency)
+    unless @new_resource.frequency == :monthly && !%w(FIRST SECOND THIRD FOURTH LAST LASTDAY).include?(@new_resource.frequency_modifier)
```
mwrock: extract to a method

providers/task.rb
```diff
@@ -271,7 +281,7 @@ def validate_create_day
   unless [:weekly, :monthly].include?(@new_resource.frequency)
     raise 'day attribute is only valid for tasks that run weekly or monthly'
   end
-  if @new_resource.day.is_a? String
+  if @new_resource.day.is_a?(String) && @new_resource.day.to_i.to_s != @new_resource.day
```
mwrock: this seems unrelated. does it fix something?

providers/task.rb
```diff
@@ -36,12 +36,22 @@
...
+               end
     frequency_modifier_allowed = [:minute, :hourly, :daily, :weekly, :monthly]
     options = {}
     options['F'] = '' if @new_resource.force || task_need_update?
     options['SC'] = schedule
-    options['MO'] = @new_resource.frequency_modifier if frequency_modifier_allowed.include?(@new_resource.frequency)
+    unless @new_resource.frequency == :monthly && !%w(FIRST SECOND THIRD FOURTH LAST LASTDAY).include?(@new_resource.frequency_modifier)
+      options['MO'] = @new_resource.frequency_modifier if frequency_modifier_allowed.include?(@new_resource.frequency)
+    end
+    options['I']  = @new_resource.frequency_modifier if @new_resource.frequency == :on_idle
```
mwrock: might be a good idea just to add an attribute `idle_time` and not overload frequency_modifier.

providers/task.rb
mwrock: oh i get it now

mwrock:
Great add. I left a few comments about breaking up the option setting code since its starting to get a bit out of hand. Also it would be cool if you could add a test or two to the task kitchen suite to validate this.

providers/task.rb
elpetak: Sounds good. I'll do it.

providers/task.rb
elpetak: Just realized it's actually /M and /MO that can't go together so I'll simplify this and it will still fit in one line as it did originally (just adding "&& @new_resource.months.nil?" to it)
This is from a manual test I did with schtasks.exe:
ERROR: /M cannot be specified with /MO.
Type "SCHTASKS /CREATE /?" for usage.

providers/task.rb
elpetak: Will do.

providers/task.rb
elpetak: I take that back... there are a few cases where both can be specified. Will extract to a method.

elpetak:
Made the suggested changes and some improvements.
I'm not sure how to do the test kitchen changes. Where should I start with that?

elpetak:
About the monthly frequency and when the frequency modifier (/MO) is not applicable, here's what I figured out after reading a few times the schtasks.exe docs and running some tests:
There are 4 variations of monthly tasks (https://technet.microsoft.com/en-us/library/cc725744.aspx)
1) Every N months: /MO (frequency_modifier) as a number, /D (day) as a number but no /M (months)
2) Specific week of the month: /MO is one of FIRST,SECOND,THIRD,FORTH,LAST, /D is one of MON-SUN, /M is a list of months.
3) Specific date each month: /MO is invalid here, /D is a number and /M is a list of months.
4) Last day of the month: /MO is LASTDAY, /D is not there and /M is a list of months.

In the attributes, the /MO (frequency_modifier) defaults to 1 so the old logic always added a /MO parameter, but for monthly tasks that have a /M (months) attribute, the /MO parameter can't be a number, it must be one of FIRST,SECOND,THIRD,FORTH,LAST, LASTDAY or not be there at all.
So that's what I changed, if the task is monthly and it has a "months" parameter, only include /MO if it's one of those. This fixes it for the third monthly case where /MO would cause schtasks.exe to throw an error.

mwrock:
It would be great if you added a task to the tasks recipe [here](https://github.com/chef-cookbooks/windows/blob/master/test/cookbooks/minimal/recipes/tasks.rb) that has a `on_idle` frequency and then a test [here](https://github.com/chef-cookbooks/windows/blob/master/test/integration/tasks/pester/minimal.tasks.Tests.ps1) that looks at the frequency and idle time of the task to ensure they are correct.

elpetak:
I added a test for on_idle. I don't have my machine setup for actually running the tests, but I did run the commands manually in powershell and I think it should work.

elpetak:
hmmm... something failed but I don't understand what... it's complaining about a gem load error with mixlib-install but I didn't change any of that :-/

mwrock:
Hmm yeah that's unrelated. I'll take care of that and get back to you.

mwrock:
ok @elpetak the break here has been fixed in master so a new push of your pr should clear that up.

elpetak:
It's looking good, isn't it? :-)

mwrock:
oh cool. yeah it LGTM :+1:

mwrock:

PR #359, opened 2016-04-13
take bundler out of the appveyor build

mwrock:
Our appveyor builds are currently breaking with gem conflicts between stuff in the bundle and the bin stubs from chefdk (see https://ci.appveyor.com/project/ChefWindowsCookbooks/windows/build/1.0.57).

This uses the chefdk exclusively and does not run bundler.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -16,7 +16,7 @@ group :unit do
 end
 
 group :kitchen_common do
-  gem 'test-kitchen', '~> 1.6'
+  gem 'test-kitchen', '~> 1.7'
 end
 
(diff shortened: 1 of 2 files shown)


iennae:
As soon as AppVeyor passes I'll pull this in. Thanks @mwrock !

smurawski:
:-1: We should be testing against the current channel of chefdk, to catch the pain before it hits the stable branch.

smurawski:
:+1: on appveyor coming back green

mwrock:
Just did a 180 turn here and changed the pr title and desc to reflect that.

PR #365, opened 2016-05-27
escape command quotes when passing to schtsks

mwrock:
Because the task provider does not escape quotes, consumers have to escape their own quotes. This not only makes the consuming experience awkward, but it can result in corrupted idempotency of the task resource. Agreat example is the `chef-client` cookbook's task recipe. It constructs the command as:
```
command "cmd /c \\\"#{node['chef_client']['ruby_bin']} #{node['chef_client']['bin']} \
  -L #{File.join(node['chef_client']['log_dir'], 'client.log')} \
  -c #{File.join(node['chef_client']['conf_dir'], 'client.rb')} -s #{node['chef_client']['splay']} > NUL 2>&1\\\""
```
because it is doing its own escaping, the idempotent check of the task provider `task_need_update?` results in an unequal command string and therefore the chef-client task recipe will converge every time.

### DIFF
--- a/providers/task.rb
+++ b/providers/task.rb
@@ -171,16 +171,18 @@ end
 
 private
 
+# rubocop:disable Style/StringLiteralsInInterpolation
 def run_schtasks(task_action, options = {})
   cmd = "schtasks /#{task_action} /TN \"#{@new_resource.task_name}\" "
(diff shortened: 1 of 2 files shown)


smurawski:
:+1:

PR #376, opened 2016-06-29
Switch to cookstyle, update gem deps and other minor stuff

tas50:
Pretty simple stuff all around broken out into individual commits


mwrock:
:+1:

PR #385, opened 2016-08-11
Add TrustedPublisher as a valid cert store_name

### LINKED ISSUE #384: TrustedPublisher is a valid windows_certificate store
### Cookbook version
v1.44.1

### Chef-client version

All

### Platform Details

Windows

### Scenario:

As a windows admin
In order to install signed software
I need to be able to place certificates in the TrustedPublisher store

### Steps to Reproduce:

```ruby
windows_certificate my_certfile do
  store_name "TrustedPublisher"
end
```

### Expected Result:

We should see the certificate in the TrustedPublisher store:

```
 dir Cert:\LocalMachine\TrustedPublisher
```

### Actual Result:

```
       Compiling Cookbooks...   ================================================================================
       Recipe Compile Error in C:/Users/vagrant/AppData/Local/Temp/kitchen/cache/cookbooks/foo/recipes/printers.rb

================================================================================
       
       Chef::Exceptions::ValidationFailed
       ----------------------------------
       Option store_name's value TrustedPublisher does not match regular expression /^(?:MY|CA|ROOT)$/
```

hh:
### Description

Ability to use windows_certificate's to the TrustedPublisher store.
### Issues Resolved

#384

### Check List
- [x] All tests pass. See https://github.com/chef-cookbooks/community_cookbook_documentation/blob/master/TESTING.MD
- [x] The CLA has been signed. See https://github.com/chef-cookbooks/community_cookbook_documentation/blob/master/CONTRIBUTING.MD

There are quite a few valid store_names:

```
> dir Cert:\LocalMachine
Name : SmartCardRoot
Name : TrustedPublisher
Name : Root
Name : Trust
Name : Remote Desktop
Name : AuthRoot
Name : CA
Name : Disallowed
Name : TrustedPeople
Name : My
Name : TrustedDevices
Name : ClientAuthIssuer
```

### DIFF
--- a/resources/certificate.rb
+++ b/resources/certificate.rb
@@ -24,5 +24,5 @@ default_action :create
 attribute :source, kind_of: String, name_attribute: true, required: true
 attribute :pfx_password, kind_of: String
 attribute :private_key_acl, kind_of: Array
-attribute :store_name, kind_of: String, default: 'MY', regex: /^(?:MY|CA|ROOT)$/
+attribute :store_name, kind_of: String, default: 'MY', regex: /^(?:MY|CA|ROOT|TrustedPublisher)$/
 attribute :user_store, kind_of: [TrueClass, FalseClass], default: false


mwrock:
I'm :+1: 

Ultimately we should just have a `requirements.assert` check to see if the given location exists but that can be a separate thing.

Stift:
Wouldn't it make sense to add all the other locations as well now?

mwrock:
I'm :-1: on hard coding a list of cert directories. It would be better to either shell out to powershell that does something like `Test-Path Cert:\some_cert_folder` or do the equivilaent via a win32 FFI call.

Stift:
I agree on that hard coded lists are not the nicest way. Currently this lwrp is pretty useless sometimes when it comes to other store types unless I hack it my way to support also other store locations.
I would recommend to remove the regex check in the ressource def and extend the documentation or add sth like

````
def cert_store_exists
  <<-EOH
Test-Path "Cert:\\#{@location}\\#{@new_resource.store_name}"
EOH
end
````

to the ressource def and check it always when needed. Is that what you would like.

**Comment by mwrock** [comment]
yeah @Stift thats pretty much what I'm saying. And you could chek it from a `define_resource_requirements`:
```
        def define_resource_requirements
          super
          requirements.assert(:create) do |a|
            a.assertion { cert_store_exists? }
            ...
          end
        end
```

tas50:
Merging this since it continues or existing setup and fixes the issue. It would be nice to fix this the real way though with an assert.

PR #388, opened 2016-08-16
Add missing matchers and update docs

tas50:
Community members have slowly been adding all our missing matchers. Let's just get them all done and update the docs. Also drop support for Chefspec <4.1. It's time.


mwrock:
:+1:

PR #391, opened 2016-09-02
Remove reboot handler / require Chef 12.1+

tas50:
Set the minimum Chef version to 12.1 since no one should be using 12.0.  Also remove the reboot handler, which is included in core chef since 12.0.


mwrock:
:+1:

PR #392, opened 2016-09-04
Don't use the package provider on Chef 12.6+

tas50:
I need @lamont-granquist / @thommay to confirm this is going to work as I assume, but here's the thought:

Stop favoring the cookbook version of the provider on Chef 12.6+. On Chef 12.6+ just use the built in package provider. Remove the Chef 11 compatibility entirely and keep 12.0-12.5.1 working as they do now.


lamont-granquist:
this reads correct to me, so 👍

PR #393, opened 2016-09-05
Update autorun to use built in registry resource

tas50:
We don't have a windows_registry provider anymore. Use the built in and add a test so we can make sure this keeps working


mwrock:
:+1:

PR #398, opened 2016-10-05
using certutil over powershell

blairham:
### Description

Converted to using certutil over powershell

### Issues Resolved

The powershell functions used in this resource do not work on windows 2012R2. This solves that problem, and also gives us the ability to converge better for create 'add', and delete

### DIFF
--- a/providers/certificate.rb
+++ b/providers/certificate.rb
@@ -31,35 +31,26 @@ def whyrun_supported?
 end
 
 action :create do
-  hash = '$cert.GetCertHashString()'
-  code_script = cert_script(true) <<
-                within_store_script { |store| store + '.Add($cert)' } <<
-                acl_script(hash)
+  if @current_resource.exists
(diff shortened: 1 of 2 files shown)


mwrock:
Hmm. Our integration tests in https://github.com/chef-cookbooks/windows/blob/master/test/cookbooks/test/recipes/certificate.rb runs on 2012R2. Can you include changes to those tests to capture the incompatibility?

blairham:
I think I might have had an old cookbook cached that was causing the `.ctor` error. As when I tried to recreate it for a long explanation, I can't know. 

The interesting thing here is I had written this code over a year ago (https://github.com/chef-cookbooks/windows/pull/167) and it didn't take off. I eventually closed the PR. 

Eitherway, this allow us to now converge properly instead of forcing it like we are now as it addresses some of the comments in the code.

PR #405, opened 2016-11-16
Add deprecation dates. Align with Chef 13

tas50:
"next major version" is pretty vague. Let's give a date. Once Chef 13 comes out the compatibility with Chef < 12.6 in this cookbook is going to be removed.


mwrock:
appveyor broke but probably not relevant to anything here but you might want to rerun it.

PR #411, opened 2016-11-23
Make sure the ohai plugin is available when installing features

tas50:
Having windows::default on your run_list should not be a requirement. Instead when we try to access the node data we should install the ohai plugin if the node data isn't there.

### DIFF
--- a/providers/feature_dism.rb
+++ b/providers/feature_dism.rb
@@ -50,6 +50,8 @@ end
 
 def installed?
   @installed ||= begin
+    install_ohai_plugin unless node['dism_features']
+
     # Compare against ohai plugin instead of costly dism run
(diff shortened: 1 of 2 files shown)


providers/feature_dism.rb
```diff
@@ -50,13 +50,17 @@ def delete_feature(_name)
 
 def installed?
   @installed ||= begin
+    install_ohai_plugin unless node['dism_features']
```
smurawski: Will the install_ohai_plugin also populate the node object? Otherwise the next line could still error looking for that element in node.

tas50: The ohai_plugin provider does a reload on the node object so it should be there after this

PR #414, opened 2016-11-23
Deprecate using attributes to specify the windows_feature provider

tas50:
We didn’t document this so I suspect not a lot of people have used it. It’s really bad form in Chef and makes refactoring this provider more difficult.


cheeseplus:
+1

lamont-granquist:
of course without landing something like #412 i'm not sure this deprecation warning is actionable.

PR #433, opened 2017-02-25
WIP: Convert LWRPs to Custom resources

tas50:
Convert the various LWRPs to custom resources and use converge_by to properly converge the state of the resources


tas50:
@smurawski I'm good to merge this if you are

PR #434, opened 2017-03-07
Adds `Source` parameter to feature_powershell provider

Jakauppila:
Adds the ability to define a source path to feature files with the windows_feature Powershell provider.

https://technet.microsoft.com/en-us/itpro/powershell/windows/server-manager/install-windowsfeature#-source


Jakauppila:
Added a check so it is only used on 8/2012+.

As far as the failing Appveyor build, I wasn't able to find a passing build with the 'feature Web-Ftp-Server and sub features installed using powershell' test, so I'm guessing it's just a faulty test.

Jakauppila:
This rebasing is really putting my Git skills to test the test!

Should be all set now on the custom resource.

PR #440, opened 2017-03-14
appveyor with `deprecations_as_errors: true`

smurawski:
Signed-off-by: Steven Murawski <steven.murawski@gmail.com>


smurawski:
The appveyor failures here are ones we want and should block the cookbook release.

tas50:
Let's get master red so we know to get it green

PR #444, opened 2017-03-14
Add back the DISM ohai plugin installation in the default recipe

tas50:
I removed this by accident when I removed all the gem installs. We need this still


tas50:
Even better

PR #445, opened 2017-03-15
More Custom Resources

smurawski:
Signed-off-by: Steven Murawski <steven.murawski@gmail.com>

Converts `windows_task`,  `certificate`, `certificate_binding`, and `http_acl` to custom resources from LWRPs.


mwrock:
Not sure how far along https://github.com/chef/chef/pull/5886 is but if its imminent, it may make sense to deprecate this resource.

tas50:
@mwrock Once we ship task in core we'll still need to keep it here for probably a year plus. We will want to get some deprecation warnings similar to what was done with the package resource when that went into core though.

resources/http_acl.rb
```diff
@@ -18,11 +18,93 @@
...
-attribute :sddl, kind_of: String
+property :url, String, name_property: true, required: true
+property :user, String
+property :sddl, String
+property :exists, [true, false], desired_state: true
 
-attr_accessor :exists
+# See https://msdn.microsoft.com/en-us/library/windows/desktop/cc307236%28v=vs.85%29.aspx for netsh info
+
+load_current_value do |desired|
```
tas50: Instead of loading the exists property we should just make this a helper method and use that method in a guard around a converge_by block. Creating properties that exist just to store state seems like a bad pattern to push out to community

resources/certificate.rb
```diff
@@ -18,11 +18,149 @@
...
+
+  guard_script = cert_script(false) <<
+                 cert_exists_script(hash)
+
+  converge_by("adding certificate #{new_resource.source} into #{new_resource.store_name} to #{cert_location}\\#{new_resource.store_name}") do
+    powershell_script new_resource.name do
+      guard_interpreter :powershell_script
+      convert_boolean_return true
+      code code_script
+      not_if guard_script
```
tas50: Is there any way to move this not_if guard outside the converge_by so we don't converge the custom resource and then not actually converge this powershell_script resource?

resources/certificate_binding.rb
```diff
@@ -18,14 +18,111 @@
...
-attribute :port, kind_of: Integer, default: 443
-attribute :app_id, kind_of: String, default: '{4dc3e181-e14b-4a21-b022-59fc669b0914}'
-attribute :store_name, kind_of: String, default: 'MY', regex: /^(?:MY|CA|ROOT)$/
+property :cert_name, String, name_property: true, required: true
+property :name_kind, Symbol, equal_to: [:hash, :subject], default: :subject
+property :address, String, default: '0.0.0.0'
+property :port, Integer, default: 443
+property :app_id, String, default: '{4dc3e181-e14b-4a21-b022-59fc669b0914}'
+property :store_name, String, default: 'MY', regex: /^(?:MY|CA|ROOT)$/
+property :exists, [true, false], desired_state: true
```
tas50: Similar to the comment on the acl provider if we can avoid having an exists property that would be great. This can be done using just a helper and converge_by with a similar amount of logic

resources/http_acl.rb
smurawski: It's also loading the existing security descriptor to check against..

I'm fine changing it, but maybe this is a wider discussion about the `load_current_value pattern` in general

resources/certificate.rb
smurawski: This whole resource needs a going over, I've got an old note on some behavior with the .net classes used inside of it.  I'd rather get the existing behavior moved over and we can iterate from there.

PR #450, opened 2017-03-18
CHANGELOG: add release date to v3.0.0

jugatsu:
### Description

As v3.0.0 was actualy released two days ago so fix release date in changelog.


tas50:
Oops. Thanks for fixing this

PR #465, opened 2017-04-06
Ensure 12.6 compatibility with action_class.class_eval

tas50:
class_eval also requires the actions to be defined first due to a bug that we’ve since fixed.

### DIFF
--- a/resources/feature_dism.rb
+++ b/resources/feature_dism.rb
@@ -25,7 +25,41 @@ property :all, [true, false], default: false
 include Chef::Mixin::ShellOut
 include Windows::Helper
 
-action_class do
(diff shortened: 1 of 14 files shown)


resources/feature_dism.rb
```diff
@@ -25,7 +25,41 @@
 include Chef::Mixin::ShellOut
```
iennae: I don't think shellout is needed to be included at this point, know it'snot a change here but just some cleanup maybe.

tas50: yep this can go away at this point

tas50:
@smurawski Agreed

PR #486, opened 2017-07-26
windows_shortcut: update example to a version that is idempotent

stuartpreston:
Signed-off-by: Stuart Preston <stuart@chef.io>

### Description

Doc fix: Fixes the example for the `windows_shortcut` resource to be idempotent. The previous example was not idempotent because the formatting of the retrieved values did not match the example properties.

### DIFF
--- a/README.md
+++ b/README.md
@@ -441,16 +441,16 @@ Creates and modifies Windows shortcuts.
 
 #### Examples
 
-Add a shortcut all users desktop:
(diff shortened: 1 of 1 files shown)


tas50:
Nice catch

Tech356:
Shouldn't this be fixed in the `windows_shortcut` resource? The resource could do a case-insensitive path comparison.

Issue #490, opened 2017-08-11
windows_share property validation incorrectly reports 'path' missing


trickyearlobe:
### Cookbook version
Fails in 3.1.1
Works in 3.1.0

### Chef-client version
12.9.38 and others

### Platform Details
Windows 2012r2 and others

### Scenario:
Using `windows_share` resource to create a share.

### Steps to Reproduce:
```
directory 'c:\testdir' do
  action :create
end

windows_share 'testdir' do
  action :create
  path 'c:\testdir'
  full_users ['everyone']
end
```

### Expected Result:
Share should be created

### Actual Result:

Validation incorrectly reports 'path' property missing with error below.
See also #482 and comments on commit fc2691ff5b126a8412a6caea2a5c99b6051ea54f

```
================================================================================
Error executing action `create` on resource 'windows_share[testdir]'
================================================================================

Chef::Exceptions::ValidationFailed
----------------------------------
path is required

Resource Declaration:
---------------------
# In C:/Users/vagrant/AppData/Local/Temp/kitchen/cache/cookbooks/windows_share_test/recipes/default.rb

 23: windows_share 'testdir' do
 24:   action :create
 25:   path 'c:\testdir'
 26:   full_users ['everyone']
 27: end

Compiled Resource:
------------------
# Declared in C:/Users/vagrant/AppData/Local/Temp/kitchen/cache/cookbooks/windows_share_test/recipes/default.rb:23:in `from_file'

windows_share("testdir") do
  action [:create]
  retries 0
  retry_delay 2
  default_guard_interpreter :default
  declared_type :windows_share
  cookbook_name "windows_share_test"
  recipe_name "default"
  path "c:\\testdir"
  full_users ["everyone"]
  share_name "testdir"
end

Platform:
---------
x64-mingw32
```

lamont-granquist:
Can you post contents of the stacktrace.out file so we can see the full stack of where the exception is getting thrown?

trickyearlobe:
@lamont-granquist here you go...

```
Generated at 2017-08-14 09:36:52 -0700
Chef::Exceptions::ValidationFailed: windows_share[testdir] (windows_share_test::default line 23) had an error: Chef::Exceptions::ValidationFailed: path is required
C:/opscode/chef/embedded/lib/ruby/gems/2.1.0/gems/chef-12.9.38-universal-mingw32/lib/chef/property.rb:368:in `get'
[...]

Issue #482, opened 2017-07-07
windows_share returning WIN32OLERuntimeError


ooOOJavaOOoo:
### Cookbook version
windows 3.1.1

### Chef-client version
12.19.36

### Platform Details
Windows Server 2012R2 local kitchen instance

### Scenario:
Creating a file share in windows using the windows_share resource

### Steps to Reproduce:
metadata.rb
```depends 'windows'```
update default.rb
```
directory 'd:\siebelfs' do
  action :create
end
windows_share 'd:\siebelfs' do
  action :create
  path 'd:\siebelfs'
  full_users [user_credential[:user]]
end
```
### Expected Result:
Expected directory be created with the share enabled

### Actual Result:
           ================================================================================
           Error executing action `create` on resource 'windows_share[d:\siebelfs]'
           ================================================================================

           WIN32OLERuntimeError
           --------------------
           (in OLE method `Count': )
        OLE error code:80041017 in SWbemObjectSet
          Invalid query
        HRESULT error code:0x80020009
          Exception occurred.

**Comment by trickyearlobe** [comment]
@ooOOJavaOOoo you can't put the drive letter in the share name.

Try this instead

```
windows_share 'siebelfs' do
  action :create
  path 'd:\siebelfs'
  full_users [user_credential[:user]]
end
```

ooOOJavaOOoo:
================================================================================
           Error executing action `create` on resource
'windows_share[siebelfs]'

 ================================================================================

           Chef::Exceptions::ValidationFailed
           ----------------------------------
           path is required

           Resource Declaration:
           ---------------------
           # In
C:/Users/ADMINI~1/AppData/Local/Temp/kitchen/cache/cookbooks/role_solar_appserver/recipes/default.rb

            47: windows_share 'siebelfs' do
            48:   action :create
            49:   path 'd:\siebelfs'
            50:   full_users ['Everyone']
            51: end
            52:

trickyearlobe:
@ooOOJavaOOoo 
Hmmm... different error now... possibly a bug in chef-client ??

In the custom resource `resources/share.rb:24` path is a property which I believe is correctly defined as `property :path, String, required: true` but it seems to be ignoring the fact you provided it. I can reproduce it on my local machine.

I can see there was a commit `fc2691f` in `3.1.1` which changed that property to `required` instead of raising an exception that it was missing. If I use the `3.1.0` version of the cookbook it converges successfully but I don't get a share created.

[...]

PR #491, opened 2017-08-14
Revert "Require path in the share resource instead of raising if it's…

trickyearlobe:
This PR reverts commit fc2691ff5b126a8412a6caea2a5c99b6051ea54f fixing #482 #490

### Description
fc2691ff5b126a8412a6caea2a5c99b6051ea54f changed the `windows_share` LWRP `path` property to `required`  instead of throwing an exception when its missing.

Despite the path being specified the chef-client run fails with 'path is required`. See #482 #490 and comments on the commit fc2691ff5b126a8412a6caea2a5c99b6051ea54f for details.

The root cause may lie in chef-client LWRP DSL, but requesting this commit to be reverted as this failure occurs with many recent chef-clients.

Signed-off-by: Richard Nixon <richard.nixon@btinternet.com>

### DIFF
--- a/resources/share.rb
+++ b/resources/share.rb
@@ -20,7 +20,7 @@
 #
 
 property :share_name, String, name_property: true
-property :path, String, required: true
+property :path, String
 property :description, String, default: ''
 property :full_users, Array, default: []
 property :change_users, Array, default: []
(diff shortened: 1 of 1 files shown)


tas50:
Thanks for tracking this down. I released the change as 3.1.2

PR #499, opened 2017-09-22
Allow fonts to be pulled from a remote location

jkburges:
Signed-off-by: Jon Burgess <jkburges@gmail.com>

### Description

Fonts can now be pulled from an http source (e.g. S3)

### Issues Resolved

It just makes it easier to install fonts sourced from a remote location.  One doesn't have to have a separate `remote_file` resource to do the job.

Also, solves a particular problem I was having when first fetching the font to the chef cache dir, and ending up with (invalid) file URIs such as "file://C:\chef\cache/fonts/Amatic-Bold.ttf was an invalid URI".

### Check List

master appears to be failing currently

I need some guidance with how to add unit tests for this new functionality

- [x] New functionality has been documented in the README if applicable

### DIFF
--- a/README.md
+++ b/README.md
@@ -231,12 +231,16 @@ Font files should be included in the cookbooks
 #### Properties
 
 - `name` - The file name of the font file name to install. The path defaults to the files/default directory of the cookbook you're calling windows_font from. Defaults to the resource name.
-- `source` - Set an alternate path to the font file.
+- `source` - Set an alternate path/URI to the font file.
 
 #### Examples
 
 ```ruby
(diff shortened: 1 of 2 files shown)


tas50:
Thanks for adding the test there as well.

jkburges:
> Thanks for adding the test there as well.

I didn't add tests :-/

See "I need some guidance with how to add unit tests for this new functionality" in OP.

tas50:
I meant the integration test. As a whole we need to sit down and add some real unit testing to this cookbook. Something like this is easy enough to do with integration testing that it makes sense to do it that way.

Issue #476, opened 2017-06-02
windows_certificate resource not idempotent on create action


scolligan:
### Cookbook version
3.1.0

### Chef-client version
13.1.31

### Platform Details
Windows Server 2012 R2 on Azure

### Scenario:
Importing a root and intermediate certificate executes on each run

### Steps to Reproduce:
```
# import root certificate
windows_certificate File.join(node['jjk_ts_base']['chef_local_certificate_path'], node['jjk_ts_transport_webserver_website_sitename']['website_root_cert_file_name']) do
  store_name 'ROOT'
  action :create
end

# import intermediate certificate
windows_certificate File.join(node['jjk_ts_base']['chef_local_certificate_path'], node['jjk_ts_transport_webserver_website_sitename']['website_intermediate_cert_file_name']) do
  store_name 'CA'
  action :create
end

```
### Expected Result:
Expected that this will execute only when the certificate is not installed.

### Actual Result:
The resources execute on each chef run:

```
  * windows_certificate[c:/jjktschef/certificates/jjk-techservices-ca.cert.pem] action create[2017-06-02T16:19:44-05:00] INFO: Processing windows_certificate[c:/jjktschef/certificates/jjk-techservices-ca.cert.pem] action create (jjk_ts_transport_webserver_website_sitename::default line 62)

    - adding certificate c:/jjktschef/certificates/jjk-techservices-ca.cert.pem into ROOT to LocalMachine\ROOT
    * powershell_script[c:/jjktschef/certificates/jjk-techservices-ca.cert.pem] action run[2017-06-02T16:19:45-05:00] INFO: Processing powershell_script[c:/jjktschef/certificates/jjk-techservices-ca.cert.pem] action run (c:/chef/cache/cookbooks/windows/resources/certificate.rb line 39)
[2017-06-02T16:19:45-05:00] INFO: Processing powershell_script[Guard resource] action run (dynamically defined)
True
[2017-06-02T16:19:45-05:00] INFO: powershell_script[Guard resource] ran successfully
 (skipped due to not_if)

  * windows_certificate[c:/jjktschef/certificates/jjk-techservices-intermediate.cert.pem] action create[2017-06-02T16:19:45-05:00] INFO: Processing windows_certificate[c:/jjktschef/certificates/jjk-techservices-intermediate.cert.pem] action create (jjk_ts_transport_webserver_website_sitename::default line 68)

[...]

PR #506, opened 2017-10-09
Fix certificate resource always reporting as updated

rlaveycal:
### Description

The certificate resource always reports as having created/deleted the certificate even if it doesn't make a change. This is because the `powershell_script` resource that does the work is wrapped by a call to `converge_by`.

### Issues Resolved

- #476

### DIFF
--- a/resources/certificate.rb
+++ b/resources/certificate.rb
@@ -35,13 +35,11 @@ action :create do
   guard_script = cert_script(false) <<
                  cert_exists_script(hash)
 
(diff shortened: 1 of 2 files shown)


rlaveycal:
AppVeyor is failing on an unrelated test suite
```
Failed to complete #converge action: [Bad HTTP response returned from server. Body(if present): (400).] on pagefile-windows-2012R2
```

tas50:
@rlaveycal We still have some unrelated test issues I'm working out. This change looks great though. Thanks for noticing and cutting a PR

PR #515, opened 2017-12-07
Fixing couple of bugs in windows_share

baurzhan-konurbayev:
### Description
This change proposes a couple of bug fixes:
- When using LIKE for comparison of user and domain, powershell script hangs. This probably uses some index rather doing full search against database. Since wildcard is not used in LIKE statement, we can better replace it with equality sign "=".
- Detection of permissions changes doesn't work due to wrong logical statement.

### Issues Resolved

- The first Chef run create share, but doesn't set permissions. We have "everyone with read access". This is because change in share's permissions cannot be detected correctly
- The second Chef run just hangs for unknown time. This is because LIKE makes full search for filtering.

### Check List

### DIFF
--- a/resources/share.rb
+++ b/resources/share.rb
@@ -230,7 +230,7 @@ action_class do
         )
         #Create the Trusteee Object
         $Trustee = ([WMIClass] "\\\\$env:computername\\root\\cimv2:Win32_Trustee").CreateInstance()
-        $account = get-wmiobject Win32_Account -filter "Name like '$Name' and Domain like '$Domain'"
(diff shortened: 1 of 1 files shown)


tas50:
@baurzhan-konurbayev Thanks for putting this together. Can you sign this for the DCO so we can accept this contribution? https://github.com/chef/chef/blob/master/CONTRIBUTING.md#developer-certification-of-origin-dco

baurzhan-konurbayev:
@tas50 Glad to contribute.  Added the signature. 

Thank you.

PR #518, opened 2017-12-29
Adds the windows_user_privilege resource which can grant privileges such as "Logon as a service"

### LINKED ISSUE #228: Add resource to manipulate Local Security Policy user rights, such as "Log on as a service"
Recently had to use powershell script in order to manipulate Local Security Policy to add a user to "Log on as a service", in order to start jenkins windows slave service provided by https://github.com/opscode-cookbooks/jenkins.

It'd be great if windows cookbook had a resource to manipulate areas of Local Security Policy directly, at least to grant user rights such as "Log on as a service"

jakauppila:
### Description

Adds a `windows_user_privilege` resource which can be used to grant various privileges like `Logon as a service` or `Logon as a batch job`

#### Oustanding
- What is the best way to incorporate a test for this? Since there's not a quick and easy way to validate with Pester
- It was mentioned in Slack that this cookbook is on the way out once pieces are integrated into Core. I could put this resource into it's own cookbook if no new additions are wanted here

### Issues Resolved

Fixes #228 

### Check List

- [x] New functionality has been documented in the README if applicable


jakauppila:
@tas50 Any thoughts on this?

resources/user_privilege.rb
```diff
@@ -0,0 +1,22 @@
+#
+# Author:: Jared Kauppila (<jared@kauppi.la>)
+# Cookbook:: windows
+# Resource:: user_privilege
+#
+
+property :principal, String, name_property: true, required: true
```
tas50: Using name_property and required together doesn't really make sense. name_property allows you to make a property and alternative to the resource's own name. If it's required it's not really an alternative and the name_property part would never actually kick in.

tas50:
@jakauppila Seems pretty straight forward. I left a comment on the name_property bit. Would it make more sense for us to just add this to chef itself thought? We're really trying to just kill off the Windows cookbook by including everything in Chef itself.

resources/user_privilege.rb
jakauppila: Sure, I'll get this updated

jakauppila:
@tas50 I would agree, but in terms of getting it out in the hands of users it'd be nice to see it here first and moved to core shortly after https://github.com/chef/chef/issues/6716 is completed where we have more complete functionality.

tas50:
Thanks for putting this together @jakauppila

jakauppila:
We may want to add a note for caution when using this resource on a Chef client prior to https://github.com/chef/chef/pull/6991 getting merged as it would occasionally cause a heap corruption will silently kills the execution.

PR #525, opened 2018-01-24
Support binding certificates to a hostname as well as an IP address

rlaveycal:
### Description

In order to support SNI on a https connection the certificate has to be bound to a hostname rather than an IP address. The `netsh` command uses a different syntax to do this.

A regex is used to determine whether the `address` property is an IP v4/v6 address or a hostname


tas50:
Thanks @rlaveycal

PR #527, opened 2018-02-17
Use Ohai to determine Powershell version

tas50:
Avoid yet another shellout when we already have this data and we've had this data since 2014.


tas50:
I think this is actually going to cause issues if powershell is installed during the run so we probably want to avoid this for now

stuartpreston:
Agreed, we will need an implicit way of testing during the run.

PR #528, opened 2018-02-17
Fix feature logging to work with arrays of features

tas50:
Also only convert the String to an array once by using coerce instead of doing it multiple times which takes more time / CPU.


stuartpreston:
Pester tests are still failing and need further investigation.

PR #530, opened 2018-02-28
Overhaul and improvements to windows_feature resource

tas50:
This PR is in preparation of moving windows_feature into Chef 14. The first step is making sure the resource is modernized and fully functional in the cookbook

## servermanagercmd.exe Support Removal

This cookbook previously supported servermanagercmd.exe, which was necessary for feature installation on Windows 2003 / 2008 (not R2) systems.  Windows 2003 went full EOL in 2015 and 2008 went into extended support both went EOL long ago and are not supported platforms for Chef so at this point we should remove the legacy code supporting servermanagercmd.

## Removal of the DISM feature caching Ohai plugin

We currently install an Ohai plugin that caches the state of features on the node, and we reload that plugin anytime we install/remove a feature from the system. This greatly speeds up Chef runs where no features are actually installed/removed (2nd run and later). Without the caching each resource would take about 1 second longer while it queried current feature state. Using Ohai to cache this data was problematic due to incompatibilities with Chef Solo, the reliance on the ohai cookbook, and the addition of extra node data which had to be stored on the Chef Server.

Instead of caching data via an Ohai plugin we just write directly to the node within the resource. This avoids the need to load in the ohai plugin and the various issues that come with that. In the end it's basically the exact same thing, but less impacting on end users and faster when the data needs to be updated.

## Fail when feature is missing
The windows_feature_dism resource had a rather un-Chef behavior in which it just warned you if a feature wasn't available on your platform and then continued on silently. This isn't how we handle missing packages in any of our package resource and because of that it's not going to be what anyone expects out of the box. If someone really wants SNMP installed and we can't install it we should fail instead of continuing on as if we did install it. So we'll now do the following things:

- When installing a feature that doesn't exist: fail
- When removing a feature that doesn't exist: continue since it is technically removed
- When deleting a feature that doesn't exist: continue since it is technically deleted

For some users, particularly those writing community cookbooks, this is going to be a breaking change. I'd highly recommend putting logic within your cookbooks to only install features on supported releases of Windows. If you'd just like it to continue even with a failure you can also use `ignore_failure true` on your resource.

## Properly support features as an array in windows_feature_dism
We claimed to support installing features as an array in the windows_feature_dism resource previously, but it didn't actually work. The actual result was a warning that the array of features wasn't available on your platform since we compared the array to available features as if it was a string. We now properly support installation as a array and we do validation on each feature in the array to make sure the features are available on your Windows release.

## Install as the default action in windows_feature_powershell

Due to some previous refactoring the :install action was not the default action for windows_feature_powershell. For all other package resources in Chef install is the default so this would likely lead to some unexpected behavior in the cookbook. This is technically a breaking change, but I suspect everyone assumed :install was always the default.


resources/feature_dism.rb
```diff
@@ -18,87 +18,157 @@
...
+    unavailable = (new_resource.feature_name - all_available)
+    raise "The Windows feature#{'s' if unavailable.count > 1} #{unavailable.join(',')} #{unavailable.count > 1 ? 'are' : 'is'} not available on this version of Windows. Run 'dism /online /Get-Features' to see the list of available feature names." unless unavailable.empty?
   end
 
-  def reload_ohai_features_plugin(take_action, feature_name)
-    ohai "Reloading Dism_Features Plugin - Action #{take_action} of feature #{feature_name}" do
-      action :reload
-      plugin 'dism_features'
+  # run dism.exe to get a list of all available features and their state
+  # and save that to the node at node.normal (same as ohai) level.
```
Annih: Hello @tas50 

Isn't ohai using `automatic` level? Why did you chose `normal`?
Would you accept a PR to pass it to something non-persistent (`automatic` of `force_override`)?

Nice feature btw.

tas50: Oh man. Giant facepalm here. That should be automatic. I'll get a fix for that in ASAP

Annih: No hurry on my side, I don't plan to upgrade the *windows* soon, but I agree this should be automatic.
Thanks for the quick answer.

tas50: Looks like you can't set automatic from within the cookbook actually. I'm just going to make these default since there's no threat of a namespace collision and it would potentially allow someone to do some magical things there is they wished to (but probably shouldn't)

tas50: Scratch that. Override it is so we can override a normal attribute set by 4.0.0

Annih: Ok so I saw you released the new version with the `override` level.
But just out of curiosity why are you saying `you can't set automatic from within the cookbook`?
I'm doing it a lot!

See https://github.com/Annih/chef-pci/blob/master/attributes/automatic.rb#L6 for instance

tas50: At least on 12.7 it was just throwing an empty hash. I'll take a peak at exactly what was going on there once I wrap up some other work.

PR #548, opened 2018-03-30
Cache data in a singleton class not on the node

tas50:
Avoid polluting the node with more junk.


tas50:
@lamont-granquist This passes the kitchen tests locally, but let me know if this is what you had in mind for the caching.

tas50:
Closing this since we removed the resource from the cookbook

PR #552, opened 2018-04-27
[In-progress] MSYS-753 - Added support of win32-certstore

piyushawasthi:
Signed-off-by: piyushawasthi <piyush.awasthi@msystechnologies.com>

### Description

Update windows_certificate resource in windows cookbook for create/delete

### Issues Resolved

[List any existing issues this PR resolves]

### Check List


libraries/win32_helper.rb
```diff
@@ -0,0 +1,40 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+require 'win32-certstore' 
+require "openssl"
+
+module Win32
```
tas50: This whole helper should just get rolled into the certificates resource. As it stands right now anyone with this cookbook on their run list will end up loading win32-certstore and openssl which is going to slow down every chef run regardless of use. Also if / when we move this resource into chef itself it will be a lot easier if it's self contained.

piyushawasthi:
@tas50, I have updated code as per your comments but still test-kitchen is failing for `win32-certstore` load issue, Because kitchen created boxes and  `win32-certstore` gem not installing into that. Please let me know how can I fixed that.

tas50:
Part of this PR is going to be to load the cert gem into Chef. We can't just do that in the Appveyor config since users will have that same issue when they use the cookbook on their nodes. I'll get some example code for you to make that happen.

test/cookbooks/test/recipes/certificate.rb
```diff
@@ -11,51 +19,109 @@
...
+#   store_name 'CA'
+#   port 444
+#   app_id '{00000000-0000-0000-0000-000000000000}'
+# end
+
+# windows_certificate_binding '443-hostname' do
+#   cert_name 'ChefDummyCertForTest'
+#   store_name 'CA'
+#   address 'www.chef.io'
+# end
```
rhass: We should delete this commented code if we do not intend to use it.

piyushawasthi: sure @rhass

piyushawasthi:
Hi @tas50, I have done with this PR. If you get the chance please have a look also please merge If all good so that we can add validate/download actions to windows_certificate in the Windows cookbook.

PR #558, opened 2018-05-22
[MSYS-754] : Added validate/download actions to windows_certificate in the Windows cookbook

piyushawasthi:
### Description

Added verify certificate and fetch certificate actions to windows_certificate in the Windows cookbook

### Verify Certificate Action
It verify certificate by thumbprint using `win32-certstore` -> `verify?` method and return certificate valid or not, In case no certificate found by given thumbprint it return `Certificate not found` message also it handle the exception if raise.

### Fetch Certificate Action
The fetch certificate action provide 2 options.
1. User can fetch certificate and display on console in `PEM` formate.
2. User can export the certificate in file on following format using property `cert_path`
- Export certificate in PEM
- Export certificate in DER
- Export certificate in CER
- Export certificate in CRT
- Export certificate in PFX with no keys
- Export certificate in P7B
- Export certificate in invalid format return error


piyushawasthi:
Please first merge PR: https://github.com/chef-cookbooks/windows/pull/552 before merging this one.

libraries/windows_helper.rb
```diff
@@ -120,6 +123,35 @@ def to_array(var)
...
+    def openssl_cert_obj
+      OpenSSL::X509::Certificate.new(raw_source)
+    end
+
+    def add_cert(cert_obj)
+      store = Win32::Certstore.open(store_name)
+      store.add(cert_obj)
+    end
+
+    def delete_cert
```
tas50: We should just move these new helper methods right into the action class of the resource. That's the way we'll use it in chef-client itself if/when we move this resource into chef and it prevents doing requires on every chef node when we don't actually use these libraries. This will save

resources/certificate.rb
```diff
@@ -65,36 +53,83 @@
...
+
+  def export_cert(cert_obj, cert_path)
+    out_file = ::File.new(cert_path, 'w+')
+    case ::File.extname(cert_path)
+    when '.pem'
+      out_file.puts(cert_obj.to_pem)
+    when '.der'
+      out_file.puts(cert_obj.to_der)
+    when '.cer'
+      cert_out = powershell_out("openssl x509 -text -inform DER -in #{cert_obj.to_pem} -outform CER").stdout
```
tas50: if this something we could do in ruby vs. shelling out?

metadata.rb
```diff
@@ -9,3 +9,4 @@
 source_url       'https://github.com/chef-cookbooks/windows'
 issues_url       'https://github.com/chef-cookbooks/windows/issues'
 chef_version     '>= 12.7' if respond_to?(:chef_version)
+gem 'win32-certstore', '= 0.1.0'
```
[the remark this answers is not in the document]
piyushawasthi: Thanks @btm for bundler doc, I have updated metadata.rb in also keep it in mind.

resources/certificate.rb
piyushawasthi: yes @tas50 we can achieve that in ruby as well but this is super easy way with less amount of code also if `openssl` command failed due to any reason powershell_out track error easily. Please let me know your thought about it.

tas50:
Merging. I'll fix the gem loading stuff. I need this in so I can convert the pester tests

PR #593, opened 2018-11-24
Fix http_acl regex to properly capture SDDL

Annih:
### Description

Former regex was incorrect and captured the trailing space and CR.
Since the SDDL format does not include spaces the new regexp is better.

### Check List

### DIFF
--- a/resources/http_acl.rb
+++ b/resources/http_acl.rb
@@ -35,7 +35,7 @@ load_current_value do |desired|
     exists true
     url desired.url
     # Checks first for sddl, because it generates user(s)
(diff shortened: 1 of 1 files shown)


Annih:
Do you think you could backport this change in the version 4 of the cookbook? 👼

tas50:
Thanks @Annih. We don't really have a process in place for backporting fixes to the cookbooks. Is there a reason you can't update to the latest?

Annih:
We are a bit late on Chef-client version, we tackled most of the breaking changes but are not yet ready to upgrade.
Don't worry that's not a huge problem if you do not backport, , we'll use a fork why we'll do our best to catchup 😄

PR #600, opened 2018-12-14
Certificate tests

xorima:
### Description
This commit adds working tests for the certificate resource, it also
removes the reuse of the same certificate which caused tests to be
unreliable

It highlights some usecases which are failing as well.

Signed-off-by: Jason Field (jason@avon-lea.co.uk)

### Issues Resolved

Does not resolve but highlights bugs in the certificate resource clearer

### Check List


Xorima:
Failing tests are: 
```
# Install PFX certificate with private key in MY store
describe powershell(private_key_script(thumbprint: '5081f667f1ef005d0ec39fa3e30aa71b4fd84eb6')) do
  its('strip') { should eq 'True' }
end

# Add (.CRT) format certificate for validation tests exists
describe powershell(thumbprint_script(thumbprint: '9E848F52575C6B1A69D6AB62E0288BFAD4A5564E')) do
  its('strip') { should eq 'True' }
end
```

tas50:
@Nimesh-Msys Here's 2 tests we should see if we can get passing in chef/chef and win32-certstore.

Nimesh-Msys:
> @Nimesh-Msys Here's 2 tests we should see if we can get passing in chef/chef and win32-certstore.

@Xorima : Apply these [changes to add a PFX certificate](https://gist.github.com/Nimesh-Msys/f9f93cbb4fb618d2e83f8ddf416560b5). Hope they would run fine then. Let me know in case of any issues.

Xorima:
@Nimesh-Msys I have made you a collaborator of my fork so you can make any changes you need to 👍

Nimesh-Msys:
> @Nimesh-Msys I have made you a collaborator of my fork so you can make any changes you need to 👍

Unable to push. Seems like I don't have access:
```
remote: Permission to Xorima/windows.git denied to Nimesh-Msys.
fatal: unable to access 'https://github.com/Xorima/windows.git/': The requested URL returned error: 403
```

Xorima:
You will need to accept the contributor invite, you should have an e-mail @Nimesh-Msys

Nimesh-Msys:
Right, so I've checked PFX import, it is running fine.
The remaining error is while importing a certificate of `.CER` format.
Its an `openssl` issue and could be resolved by `renaming the file to .pem`. Here is [a similar discussion](https://stackoverflow.com/questions/14191468/openssl-encoding-errors-while-converting-cer-to-pem).
In near future, we are planning to use purely Crypto APIs and thus remove the usage of openssl utility. It is under discussion. But as of now, you can skip this case.

Nimesh-Msys:
`CER` formats are working fine. `PFX` should also run, but the error is unexpected. Will check tomorrow.

Nimesh-Msys:
Hey @Xorima , PR is fixed now. I made the following changes:

- Fixes to properly use `CRT` format certificates.
  - This change also needs to be implemented in core. I'll do that later
- Using `add_pfx` from `win32-certstore-0.2.4 ` to import PFX certificates.
- Corrected thumbprint of a certificate.

Xorima:
may need to also swap back to chef 13? I think I bumped it at some point, 

at least we have great coverage going forwards as well :)

Nimesh-Msys:
> may need to also swap back to chef 13? I think I bumped it at some point,
> 
> at least we have great coverage going forwards as well :)

Yes of course, it was a nice coverage. Specially the one with base 64 encoded CRT certificate. I think that was the only case I missed while writing down [these test cases](https://github.com/chef/chef/blob/master/spec/functional/resource/windows_certificate_spec.rb). 

Reason for back porting: Actually `chef_version_for_provides` seems to be slightly confusing for me here. Not sure, whether it checks the version after which resources from chef core resources should be run. Like introduction of windows_certificate in chef-core ( `> 15.0.16` ). But on running it on current chef-client(15.0.158), it raises the error:
```
Resource windows_certificate from a cookbook is overriding the resource from the client. Please upgrade your cookbook or remove the cookbook from your run_list before the next major release of Chef.
```
Or it denotes the version until this resource file will run, otherwise core (`> 15.0.16`). Again, running it on current chef-client(15.0.158) I checked that core resource was running.

Or if nothing but simply informs that until this version, we there will be [no deprecation warning errors](https://github.com/chef/chef/blob/c7d9809deee1f5ebc5c7ba8ec5343f026d60698c/lib/chef/resource.rb#L1313-L1327).

That is why I choose to run these tests on chef 13 and everything worked well.

Nimesh-Msys:
@Xorima , To remain in sync with core changes, i've implemented `binary_cert?` method here as well.

PR #609, opened 2019-04-02
Grant-SmbShareAccess - remove double quotes for -AccountName parameter

Vit0hA:
### Description

AccountName parameter requires String[], double quotes transform it to usual String and list of accounts couldn't be added to acl. Currently only one account could be added during run.

https://docs.microsoft.com/en-us/powershell/module/smbshare/grant-smbshareaccess?view=win10-ps#optional-parameters

### Issues Resolved

Haven't found related issues.

### Check List

obvious fix

### DIFF
--- a/resources/share.rb
+++ b/resources/share.rb
@@ -243,7 +243,7 @@ action_class do
       # set permissions for a brand new share OR
       # update permissions if the current state and desired state differ
       next unless permissions_need_update?(perm_type)
-      grant_command = "Grant-SmbShareAccess -Name '#{new_resource.share_name}' -AccountName \"#{new_resource.send("#{perm_type}_users").join('","')}\" -Force -AccessRight #{perm_type}"
+      grant_command = "Grant-SmbShareAccess -Name '#{new_resource.share_name}' -AccountName #{new_resource.send("#{perm_type}_users").join('","')} -Force -AccessRight #{perm_type}"
 
       Chef::Log.debug("Running '#{grant_command}' to update the share permissions")
       powershell_out!(grant_command)


tas50:
Pinging @Xorima

tas50:
Also @Vit0hA can you sign off this commit for the DCO so we can accept it. The information is here: https://github.com/chef/chef/blob/master/CONTRIBUTING.md#developer-certification-of-origin-dco

resources/share.rb
```diff
@@ -243,7 +243,7 @@ def update_permissions
       # set permissions for a brand new share OR
       # update permissions if the current state and desired state differ
       next unless permissions_need_update?(perm_type)
-      grant_command = "Grant-SmbShareAccess -Name '#{new_resource.share_name}' -AccountName \"#{new_resource.send("#{perm_type}_users").join('","')}\" -Force -AccessRight #{perm_type}"
+      grant_command = "Grant-SmbShareAccess -Name '#{new_resource.share_name}' -AccountName #{new_resource.send("#{perm_type}_users").join('","')} -Force -AccessRight #{perm_type}"
```
Xorima: The problem we have now is we get a value of 
`Grant-SmbShareAccess -Name 'read_only' -AccountName BUILTIN\Users","foo","bar`

Based on microsoft documentation we need it in this format:
> Specifies the name of the trustee. Use a comma-separated list to specify multiple trustees.
`Grant-SmbShareAccess -Name 'read_only' -AccountName "BUILTIN\Users","foo","bar"`

Is this not what the original code was doing?

Vit0hA: Sorry for my inattention, it's a bug with my array which i passed to mentioned resource.

PR #611, opened 2019-04-23
Remove deprecated Win32API

jaymalasinha:
Signed-off-by: Jaymala Sinha <jsinha@chef.io>

### Description

Remove the deprecated Win32API as it is now included in chef.
Update to chef client version >=14

### Check List

- [x] All tests pass. See <https://github.com/chef-cookbooks/community_cookbook_documentation/blob/master/TESTING.MD>

### DIFF
--- a/libraries/version.rb
+++ b/libraries/version.rb
@@ -20,7 +20,6 @@
 
 if RUBY_PLATFORM =~ /mswin|mingw32|windows/
   require_relative 'wmi_helper'
-  require 'Win32API'
 end
 
 module Windows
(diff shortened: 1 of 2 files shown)


libraries/version.rb
```diff
@@ -186,10 +183,9 @@ def initialize
 
     private
 
-    # Win32API call to GetSystemMetrics(SM_SERVERR2)
     # returns: The build number if the system is Windows Server 2003 R2; otherwise, 0.
     def sm_serverr2
```
tas50: We can probably just remove this method since we haven't support 2003 for a long time.

jaymalasinha: removed

libraries/windows_helper.rb
```diff
@@ -86,17 +84,6 @@ def cached_file(source, checksum = nil, windows_path = true)
       end
     end
 
-    # Expands the environment variables
-    def expand_env_vars(path)
```
tas50: We probably want this still. I'm sure it's used somewhere.

metadata.rb
```diff
@@ -4,7 +4,7 @@
 license          'Apache-2.0'
 description      'Provides a set of useful Windows-specific primitives.'
 long_description IO.read(File.join(File.dirname(__FILE__), 'README.md'))
-version          '5.3.0'
+version          '5.3.1'
```
tas50: Our tooling bumps the versions so this isn't necessary

libraries/windows_helper.rb
jaymalasinha: Ref: https://github.com/chef/chef/commit/561c06d451427d444217f6d901a16f30144a6903#diff-6f1a02d1cc16201c682598cb413ebd42L52

tas50:
Thanks @jaymalasinha

tas50:
Released as 5.3.1

PR #624, opened 2020-03-26
Removed windows_share and windows_certificate resources

### LINKED ISSUE #623: Remove windows_share and windows_certificate resources
https://github.com/chef-cookbooks/windows#deprecated-resources-note

We moved `windows_share` and `windows_certificate` into core for Chef 14.7. We had announced we would remove those from this cookbook late last year but didn't get to it. Let's remove them now and ship a v7 of this cookbook. 

Chef 14 will be EOL in 6 weeks. Probably we  should bump the minimum version of Chef to 15 too.

This also will allow us to remove `win32-certstore` as a gem dependency on this cookbook.

kapilchouhan99:
Signed-off-by: Kapil Chouhan <kapil.chouhan@msystechnologies.com>

### Description
- Removed `windows_share` and `windows_certificate` resources from the `windows` cookbook

### Issues Resolved
Fixes: #623


Xorima:
I think the metadata should get bumped at the same time to the next major version?

This would mean removing the path helper as well though

